PR #123, opened 2022-07-11
Matlab bindings

rubiop:
Here are the matlab bindings! 

Everyone should be able to run the examples in examples/matlab. There is map from regression, density 1D, density 2D, and samples 1D and 2D. 

There is one error I need to fix is the binding of the inverse method but I was hoping we can already merge that and open an issue for the inverse.


mparno:
I made a few minor changes to the interface to make sure we can do things in matlab exactly the same way we do things in c++ or python.  For example, you can now use the following matlab code:

```matlab
KokkosInitialize(8)
opts = MapOptions();

mset = MultiIndexSet(3,3);
fset = mset.Fix();

comp = CreateComponent(mset,opts);

S = CreateTriangular(2,1,4,opts);
```

mparno:
@rubiop Do you agree that this closes #141  and #124 ?

mparno (approved):
Approving, but we still need to make sure the bindings are complete for each class.  MultiIndexSet is only partially complete for example.

rubiop:
@mparno yes it closes #141 and #124 , thanks for cleaning everythin

PR #127, opened 2022-07-11
Added ability to turn tests on and off 

mparno:
When configuring the MParT build with cmake, adding the `-DMPART_BUILD_TESTS=OFF` will now prevent the tests from being installed.  This will also cause CMake to skip looking for, or clone, the Catch2 test library.


rubiop (approved):
Works great!

PR #129, opened 2022-07-11
Changed python bindings to use structure-based Kokkos init.

mparno:
According to the [Kokkos docs](https://github.com/kokkos/kokkos/wiki/Initialization), the Kokkos configuration can be set at runtime by constructing an instance of the`Kokkos::InitArguments` class.  Previously we were just passing in arguments like they were set on the command line, but the Kokkos documentation seems to imply that if negative values are passed to `Kokkos::InitArguments`, then the default behavior is to use `hwloc` to find the default number of threads.   

Also added ability to check the amount of concurrency that Kokkos can exploit (e.g., the number of threads) using `mpart.Concurrency()` in python.

### DIFF
--- a/bindings/python/package/__init__.py
+++ b/bindings/python/package/__init__.py
@@ -1,5 +1,5 @@
 import sys
 from .pympart import *
 
-kokkos_init = Initialize({})
+kokkos_init = Initialize(dict())
 sys.modules[__name__] = sys.modules['mpart']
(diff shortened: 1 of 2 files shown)


rubiop (approved):

mparno:
In order for Kokkos to figure out how many threads it can use, you must have the `hwloc` library installed before building Kokkos and `Kokkos_ENABLE_HWLOC` must be set to true when configuring Kokkos with CMake.   hwloc can be installed with conda: `conda install -c conda-forge libhwloc`


If Kokkos was compiled with `hwloc` support, the `PrintKokkosInfo` executable will print:
```
KOKKOS_ENABLE_HWLOC: yes
```

PR #143, opened 2022-07-14
Changed name of`at` binding to `__getitem__`

michael-c-brennan:


mparno:
Looks fine overall.  I would suggest keeping the `at` function in the bindings even though it duplicates `__getitem__` so that the interface could be identical.  If someone sees an example in c++, it would be nice if they could almost copy/paste to python.

michael-c-brennan:
@mparno Sounds good! I added `at` back.

mparno (approved):

PR #146, opened 2022-07-15
Converted entire map stack to use Strided layouts.

### LINKED ISSUE #71: Build non copy conversion between mxarray and Kokkos view (matlab binding)
Create a 'from' function that converts matlab inputs to Kokkos views without copying it

mparno:
Closes #71 

Previously, we were using Kokkos views that were restricted to the default Kokkos memory layout (row-major on host code or column-major in device code).   When constructing arrays, I still use these defaults, but I updated the function definitions in `ConditionalMapBase` and `ParameterizedFunctionBase` to accepted general strided layouts.   This enables either row major OR column major formats to be used.  Strided layouts can also be used to grab "slices" or matrices without making copies.

One immediate benefit of this update is that we can now natively support the memory layouts of any other language, such as Matlab.   So we no longer need to copy memory when interfacing between matlab and MParT.

@rubiop I updated the Matlab bindings, but could you rerun your speed tests to verify that Matlab performance is now more in line with the other languages?   Thanks!

I used this code for testing in matlab:
```matlab
clear 
close all

addpath('~/Installations/MParT/matlab/')
addpath('~/Installations/MParT/matlab/MapOptions')

KokkosInitialize(8)
opts = MapOptions();

%% Create Map
dim = 2;
totalOrder = 2;
triMap = CreateTriangular(dim,dim,totalOrder,opts);

%% Generate Test Samples
nsamp = 10000;
X = randn(dim,nsamp);

%% Test functions
tic
log_det = triMap.LogDeterminant(X);
toc

tic 
evals = triMap.Evaluate(X);
toc 

tic 
X2 = triMap.Inverse(X,evals);
toc 

tic 
detGrad = triMap.LogDeterminantCoeffGrad(X);
toc 

tic 
sens = ones(dim,nsamp);
coeffGrad = triMap.CoeffGrad(X,sens);
toc 
```

### DIFF
--- a/MParT/ConditionalMapBase.h
+++ b/MParT/ConditionalMapBase.h
@@ -62,7 +62,7 @@ namespace mpart {
         virtual Eigen::VectorXd LogDeterminant(Eigen::Ref<const Eigen::RowMatrixXd> const& pts);
 
         virtual void LogDeterminantImpl(StridedMatrix<const double, MemorySpace> const& pts,
-                                        Kokkos::View<double*, MemorySpace>              output) = 0;
+                                        StridedVector<double, MemorySpace>              output) = 0;
 
 
         /** Returns the value of \f$x_2\f$ given \f$x_1\f$ and \f$r\f$.   Note that the \f$x1\f$ view may contain more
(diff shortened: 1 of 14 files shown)


rubiop (approved):
Actually speed up matlab computations and put them at same level as python

PR #157, opened 2022-07-19
Add Julia Bindings

dannys4:
I have all the python from the MParT-examples repository work (see branch `dannys4/julia-examples` there), so I think I want to open up this PR and get it merged asap. I'm going to cull through and try to take out extraneous code as well as add a little documentation, but the meat of this PR is completely done.

Probably just try out executing each of those examples and make sure they give what you would expect. I have an option `make_plot` you have to enable for plotting in those examples (it takes a bit to get to the first plot in julia, so I didn't want to have that show if you just want to make sure commands execute).


dannys4:
Some known quirks that I still want to work out--
- The `Initialize` happens at `include` time in Julia, and I think I bound the wrong `Initialize` function, so it *will* segfault if you try to use `include("MParT.jl")` more than once.
- Enums in Julia are not my favorite thing, because they actually get exposed as names if you export them (i.e., I would have to `export` each enum for each type one by one, and then they would clutter the `Main` namespace when you call `using .MParT`). The par for course on this would be to put each group of enums in separate submodules, which I can do in CxxWrap, just haven't taken the time
- Obviously, the wrapping function should be split up into a bunch of files like we did in Python
- not all bound functions are exported
I don't really think any of these are dealbreakers, and I'm planning on just making a follow-up PR fixing all of these little things.

mparno (approved):
Everything works for me!  Thanks @dannys4 !

PR #160, opened 2022-07-19
Completed matlab-bindings (hopefully)

rubiop:
Following #149 I have completed matlab bindings for:
- `MultiIndex`
- `MultiIndexSet`
- `ConditionalMapBase`
- `TriangularMap`
- `ParameterizedFunctionBase`

Some comments:

- Common to `MultiIndex`, `MultiIndexSet` and `TriangularMap`: In order to be consistent with matlab ordering (starting from 1) I have changed how to get in the list of indices, multiIndices and map components. For example in matlab we will have `multi.Get(1)` to get first index of multi, `mset.Expand(1)` to expand the first multiIndex and `triMap.getComponent(1)` to get first component instead of `multi.Get(0)`, `mset.Expand(0)` and `triMap.getComponent(0)` in C++. I can reverse this is we want exactly the same syntax than C++ but I assumed that this made more sense for matlab users.
- `MultiIndexSet`: Everything is bound except function using limiters (and so `CreateTensorProduct`)
- Also `MultiIndexSet`: Oveload of `[]` operator in C++ has been replaced by overload of operator `{}` same thing with `!=` which has been replaced by `~=`.
- `MultiIndexSet` (unknown error): I have commented the function `MultiIndexSet_ExpandAny` in `MultiIndexSet_mex.cpp` because of some weird C++ error. If you can look at it @mparno, that would be great.

I have added a tests folder in the matlab bindings folder to test out methods of classes. They are not proper tests but just a list of function evaluations for each class just to test that bindings are working in terms of inputs/outputs (not that outputs are consistent).


mparno (approved):
@rubiop I fixed the issues you mentioned.  It turned out that the `Expand()` function had never been implemented in `MultiIndexSet`.  I also removed the include paths from the tests, since this might depend on a user's particular installation folder.

rubiop:
Sounds good @mparno , thanks!

PR #165, opened 2022-07-22
Added handling of other quadrature options.

### LINKED ISSUE #97: Add `ClenshawCurtis` and `AdaptiveClenshawCurtis` quadrature to`MapFactory`

mparno:
Closes #97 

- Added additional branches to MapFactory::CreateComponent if statement to process `ClenshawCurtis` and `AdaptiveClenshawCurtis` quadrature rules.
- Added test to make sure we can build/evaluation components with these rules.

This crazy nested if statement is a clunky way of handling these options, but I'm not sure if there is a better way with the templating we're doing.


rubiop (approved):
Looks good to me

PR #171, opened 2022-07-26
147 render notebooks into sphinx documentation

bonh:
Looks like we have to fetch the python examples from `MparT-examples` during the execution of the github action.

Is there any way I can test this before it gets merged?

Any other ideas?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -39,7 +39,7 @@ set(CMAKE_CXX_FLAGS_RELEASE "-O3")
 option(BUILD_SHARED_LIBS "Build using shared libraries" ON)
 
 ##############################################################
-# RPATH settings 
+# RPATH settings
 
 # use, i.e. don't skip the full RPATH for the build tree
 set(CMAKE_SKIP_BUILD_RPATH FALSE)
@@ -108,7 +108,6 @@ if(MPART_JULIA)
   endif()
   execute_process(COMMAND ${JULIA_EXE} -e "using CxxWrap; print(CxxWrap.prefix_path())" OUTPUT_VARIABLE cxxwrap_location)
   list(APPEND CMAKE_PREFIX_PATH "${cxxwrap_location}/lib/cmake/JlCxx/" )
-
   find_package(JlCxx)
 
   if(NOT JlCxx_FOUND)
@@ -119,7 +118,7 @@ if(MPART_JULIA)
     get_filename_component(JlCxx_location ${JlCxx_location} DIRECTORY)
     get_filename_component(JlCxx_src ${JlCxx_location} DIRECTORY)
     set(CMAKE_INSTALL_RPATH "${CMAKE_INSTALL_PREFIX}/lib;${JlCxx_location}")
-    include_directories(${Julia_INCLUDE_DIRS} ${JlCxx_src}/include)
+    include_directories(${JlCxx_src}/include)
   endif()
 
 endif()
--- a/docs/source/installation.rst
+++ b/docs/source/installation.rst
@@ -106,21 +106,14 @@ If you have Julia installed, but CMake was not able to find it during MParT conf
 
 To prevent the Julia bindings from being compiled, even if Julia and CxxWrap are found, set :code:`MPART_JULIA=OFF` during the CMake configuration.
 
-Once MParT is installed with Julia bindings (i.e. :code:`MPART_JULIA=ON`) into :code:`<your MParT install path>/julia/`, you can easily use :code:`mpart` in Julia by first adding the path to your library path
+Once MParT is installed with Julia bindings (i.e. :code:`MPART_JULIA=ON`) into :code:`your/MParT/install/path`, you can using MParT in Julia with a few last steps. First, add :code:`MParT.jl`, which holds the Julia interface for MParT, via :code:`]add https://github.com/MeasureTransport/MParT.jl` in the Julia REPL. Then, create a file :code:`~/.julia/artifacts/Override.toml` with the following lines
 
-.. tabbed:: MacOS
+.. code-block:: toml
 
-    .. code-block:: bash
+    [bee5971c-294f-5168-9fcd-9fb3c811d495]
+    MParT = "your/MParT/install/path"
 
-        $ export DYLD_LIBRARY_PATH=<your MParT install path>/julia/mpart/:$DYLD_LIBRARY_PATH
-        $ export JULIA_LOAD_PATH="<your MParT install path>/julia/mpart/:$JULIA_LOAD_PATH"
-
-.. tabbed:: Linux
-
-    .. code-block:: bash
-
-        $ export LD_LIBRARY_PATH=<your MParT install path>/julia/mpart/:$LD_LIBRARY_PATH
-        $ export JULIA_LOAD_PATH="<your MParT install path>/julia/mpart/:$JULIA_LOAD_PATH"
+At this point, you should be able to open up a REPL and type :code:`using MParT` and get going with any of the provided examples!
 
 Compiling with CUDA Support
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^


mparno:
Thanks for working on this @bonh !   I've been playing around a bit with how to pull things from the `MParT-examples` repository when building the documentation (ideally not just in the github workflow, but locally as well).   Unfortunately, I think I found a complication to working with `jupytext`.   On the website, we'd ideally like to show the output cells of the notebook as well, but because we're using python files instead of `ipynb` files, we need to evaluate the notebook to produce the output.   `nbsphinx` will do this, but it requires MParT to be installed.    I see a few ways around this, but wanted to get your opinion:

1. We include copies of the examples as prerun notebooks in the main MParT repository, which would simplify pulling the notebooks into sphinx and would allow us to use the precomputed output cells.
2. After finishing the conda package, we install a version of MParT during the github workflow so that the output cells can be populated during the documentation build.   This might eat up our github actions minutes for more elaborate examples that take a few minutes to run, but might be ok.
3. A hybrid between 1 and 2.  We automatically generate and commit the notebooks to the main MParT repository by adding another step to our current test workflow that creates and runs the notebooks after compiling MParT from source.   This could be done at pull-request time.  The docs recipe could then include the notebooks directly.

What do you think?   I'm leaning towards option 3, but am curious if you, or anyone else on the team, had a preference.

bonh:
Thanks for the suggestions. Here are my thoughts:

> 1. We include copies of the examples as prerun notebooks in the main MParT repository, which would simplify pulling the notebooks into sphinx and would allow us to use the precomputed output cells.

That would require copying, running and uploading the notebooks by hand, right? So we would store the same information at two different places. And we would put *.ipynb files under version control. That doesn't sound appealing to me. 

> 2. After finishing the conda package, we install a version of MParT during the github workflow so that the output cells can be populated during the documentation build.   This might eat up our github actions minutes for more elaborate examples that take a few minutes to run, but might be ok.

I like that one the most because the examples are still only stored in one location (`MparT-examples`) and therefore always in sync with the documentation (or the other way around). But if resources are limited that wouldn't work. [But are github actions really limited that much?](https://docs.github.com/en/actions/learn-github-actions/usage-limits-billing-and-administration#usage-limits) It says that "GitHub Actions usage is free for both public repositories and self-hosted runners. For private repositories, each GitHub account receives a certain amount of free minutes and storage, depending on the product used with the account." 

> 3. A hybrid between 1 and 2.  We automatically generate and commit the notebooks to the main MParT repository by adding another step to our current test workflow that creates and runs the notebooks after compiling MParT from source.   This could be done at pull-request time.  The docs recipe could then include the notebooks directly.

This would at least remove the manual part and we could enforce the example scripts in `MparT-examples` and notebooks in `MparT` to be in sync. But still, we would version control the notebooks in `MparT` (right?) and double the information.

In summary, I like the second one the most if the resources are not that limited.

bonh:
@mparno  Sounds great, thanks!

bonh:
It might be possible to move the compilation into a Docker container: https://docs.github.com/en/actions/creating-actions/creating-a-docker-container-action

In that case, we could build a Docker container containing `MparT` and `MparT-examples` which gets automatically built and deployed on push to `main`. I think the binder would work for that too. That action constructs a docker from the repo and pushes it to a public hub: https://github.com/jupyterhub/repo2docker-action.

mparno:
@bonh This new approach seems to be working for me locally, but sphinx is taking a long time to run on github actions.  (See [this run](https://github.com/MeasureTransport/MParT/runs/7547930002?check_suite_focus=true) for example.).    Could you test `make sphinx` locally on your machine to make sure things are working for you as well?   Note that you now have to set the `MPART_EXAMPLES_DIR` CMake variable to point towards the `MParT-examples` repository.  For example, this is my configuration:
```bash
cmake \
  -DCMAKE_INSTALL_PREFIX=~/Installations/MParT \
  -DPYTHON_EXECUTABLE=`which python` \
  -DMPART_EXAMPLES_DIR=~/Documents/Repositories/MeasureTransport/MParT-examples \
..
```

bonh:
```
CMake Error at CMakeLists.txt:252 (file):
  file COPY cannot find "PATH/TO/MparT-examples": No
  such file or directory.
```
But I have verified that `PATH/TO`MparT-examples` is correct.

bonh:
Ups, wait, typo

bonh:
Okay, works locally. Let's see what happens in the github action...

bonh:
It's stuck at `reading sources... [  2%] conf`.

.github/workflows/build-doc.yml
```diff
@@ -36,7 +36,7 @@ jobs:
         
       - name: Install sphinx extensions
         shell: bash -l {0}
-        run: conda install -c conda-forge pygments sphinx sphinx-design breathe pydata-sphinx-theme nbsphinx jupytext pandoc ipykernel nbconvert
+        run: conda install -c conda-forge pygments sphinx sphinx-design breathe pydata-sphinx-theme nbsphinx jupytext pandoc ipykernel nbconvert sphinx-panels
```
mparno, line 39: @bonh `sphinx-panels` has been deprecated and should now only be using `sphinx-design`.   Did I miss removing sphinx-panels somewhere?

bonh: Ah, I didn't know that. Yes, some of the API documentation seems to use panels.

bonh: But these errors weren't causing the long building time so we can revert that.

bonh:
Perhaps that might help: https://www.sphinx-doc.org/en/master/usage/extensions/duration.html

bonh:
So it seems that the execution of the notebooks takes way to long. On my laptop, it takes only a few seconds per notebook.

bonh:
yea, the execution of the notebooks is most likely the problem.

bonh:
So it is really the execution of the notebook, see https://github.com/MeasureTransport/MParT/runs/7565660346. Next step is to execute the pure python script and see if that is faster.

bonh:
Executing the python script is fast so now we know that it is `nbconvert` which takes so long.

bonh:
Mhh, it seems as if `nbconvert` does not execute the cells...

mparno:
Here's a few options that I can think of.  The short term option is just to hold us over until we can figure out a long term solution.

**Short term option:** 
Turn off the `build-doc` workflow.  Build documentation locally and manually update to `gh-pages` branch with the built documentation.

**Long term option:**
Use a more powerful self-hosted runner with github-actions that can actually run the notebooks.   For example, we could follow the approach outlined [here ](https://github.blog/2020-08-04-github-actions-self-hosted-runners-on-google-cloud/)for google cloud platform.  If there is a small fee, we could probably find the money to support this.

bonh:
> **Short term option:** Turn off the `build-doc` workflow. Build documentation locally and manually update to `gh-pages` branch with the built documentation.

yes, we should do that. 

> **Long term option:** Use a more powerful self-hosted runner with github-actions that can actually run the notebooks. For example, we could follow the approach outlined [here ](https://github.blog/2020-08-04-github-actions-self-hosted-runners-on-google-cloud/)for google cloud platform. If there is a small fee, we could probably find the money to support this.

Or we try
> It might be possible to move the compilation into a Docker container: https://docs.github.com/en/actions/creating-actions/creating-a-docker-container-action

see my comment https://github.com/MeasureTransport/MParT/pull/171#issuecomment-1197110650.

bonh:
Executing a notebook via `nbconvert` using a docker build with a github action works. No performance issue, see https://pipelines.actions.githubusercontent.com/serviceHosts/dfb6cfe2-bc99-4dc4-b279-7711e18748de/_apis/pipelines/1/runs/3/signedlogcontent/2?urlExpires=2022-08-03T02%3A33%3A03.8567601Z&urlSigningMethod=HMACV1&urlSignature=LJxxA3MqDS%2BMaUvhfTcQPVsNmn9EbcYN%2Beeg4S6hiX8%3D. 

So now we just have to use that docker to execute all the notebooks and fetch the results while building the documentation.

bonh:
I will try to use docker now. To keep the history clean I am resetting the branch "147-render-notebooks-into-sphinx-documentation". 

NOTE: the old commits of this branch can be found in "147-render-notebooks-into-sphinx-documentation_FIRSTTRY".

PR #173, opened 2022-07-26
added how to build the documentation to development

bonh:


dannys4 (approved):
Had to add a few dashes for the underline due to compile error, otherwise looks good

PR #174, opened 2022-07-26
add binder batch to README

bonh:
Do we want this?


dannys4:
I vote yea, it would be a great addition.

rubiop (approved):
Yes!

PR #175, opened 2022-07-26
Fix issues with Julia

dannys4:
@rubiop pointed out some issues with the Julia bindings in the Julia docs and lack of support for kokkos threads. Now you should be able to do, e.g., 
```
$ export KOKKOS_NUM_THREADS=8
$ julia
...
julia> using MParT
```
and it will automatically start up with 8 threads (it should say so).


rubiop (approved):
Working nicely!

PR #177, opened 2022-07-27
Removed the use of git submodules.

### LINKED ISSUE #163: Update CMake dependency handling
- [x] Convert from git submodules to ~ExternalProject~ `FetchContent`
- [x] Add flag to prevent dependencies from being installed if they are not found.

mparno:
Closes #163 

- Remove the use of git submodules in favor of using CMake's `FetchContent` function.
- Added a CMake option `MPART_FETCH_DEPS` that allows users to force MParT to use external installations of packages.  This options defaults to `ON`, but if `OFF`, an error will be thrown if a dependency cannot be found.


bonh:
with `cmake -DCMAKE_INSTALL_PREFIX=$HOME/local -DPYTHON_EXECUTABLE="which python" -DMPART_FETCH_DEPS=OFF  ..` I get
```
-- Found Kokkos!
-- Found Eigen!
CMake Error at CMakeLists.txt:122 (message):
  Could not find Eigen3 library and MPART_FETCH_DEPS=OFF, so CMake will not
  attempt to fetch and install Eigen3 itself.
```
Now, I look into `CMakeLists.txt` at line 122 and think that the messages `Found Eigen!` and `Could not find Eigen3...` are mutually exclusive, no?!

mparno:
> Now, I look into CMakeLists.txt at line 122 and think that the messages Found Eigen! and Could not find Eigen3... are mutually exclusive?!

Good catch @bonh !   The message on line 122 should actually be referring to pybind11, not eigen3.  There was a typo in there.   I've fixed this in the latest commit.

bonh (commented):
Awesome, thanks!

bonh (approved):

dannys4:
@bonh can we merge this?

PR #185, opened 2022-08-05
Initial GPU bindings for python

dannys4:
I am cheating a little bit here-- everytime anything gets executed for the GPU, all the data has to be piped to the GPU and then piped back afterward. This is not the best way of doing it and I will certainly change it in the future, but it's a first step and something that would work for multiple languages.

Before merging, I'm going for
- [x] `ConditionalMapBase` support
- [x] `ParameterizedFunctionBase` support
- [x] `TriangularMap` support
- [x] testing all of these features on one of the MParT examples

After this PR, I plan to make some wrapper like `DeviceMatrix` that holds a `StridedMatrix` that you can't access through the binding, but just exists on the device, and expose a way to bring that back to host


mparno:
See [MPart-examples PR #39](https://github.com/MeasureTransport/MParT-examples/pull/39) for example usage.  "Device" versions of functions and classes are simply prepended with a `d`.  For example, to create a CUDA-enable triangular map, you would call the `dCreateTriangular` function instead of the `CreateTriangular` function.

mparno (approved):

PR #200, opened 2022-08-16
Attempted fix of FindJulia.cmake when Julia is not present.

mparno:
@rubiop and @dannys4 , Could you verify that these changes to CMake still work for each of you?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -130,29 +130,36 @@ endif()
 
 # Add Julia if necessary
 if(MPART_JULIA)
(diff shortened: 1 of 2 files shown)


rubiop (commented):
@mparno solve the issue I add and works fine!

dannys4:
I believe this works for me! I'm not sure if it's this PR, a previous PR, or just my computer, but I am noticing that it's a little slow to compile.

dannys4 (approved):

PR #201, opened 2022-08-16
Separated python installation location.

mparno:
I added an additional CMake option, `PYTHON_INSTALL_PREFIX` that allows the conda-recipe to install the python package into a different location than the default `CMAKE_INSTALL_PREFIX/python`.   If `PYTHON_INSTALL_PREFIX` is not set, than everything behaves as it did before.


rubiop (approved):
Works for me in the same way than before

PR #204, opened 2022-08-18
Create component from a FixedMultiIndexSet

rubiop:
Before this update the map component could be created from a `MultiIndexSet` (which is automatically fixed when calling the map component constructor in C++).

Now you can directly use a `FixedMultiIndexSet`. 

Maybe we should remove the ability of doing it directly from a `MultiIndexSet`.


mparno (approved):
Looks good to me.   I think we can keep both options.

PR #208, opened 2022-08-19
Fixes needed for compile and run speed tests with CUDA backend.

mparno:


mparno:
@dannys4 I fixed a few additional things that made a huge performance boost.  Could you try compiling on your machine when you have a chance to see if it works for you now?

PR #209, opened 2022-08-21
Fix consistency of indexing in Matlab bindings

rubiop:
Fixed indexing in the inputs and outputs of few MultiIndexSet matlab bindings

### DIFF
--- a/MParT/MultiIndices/MultiIndex.h
+++ b/MParT/MultiIndices/MultiIndex.h
@@ -167,7 +167,6 @@ public:
     */
     bool operator<=(const MultiIndex &b) const;
 
-
 private:
 
     unsigned int length;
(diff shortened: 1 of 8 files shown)


rubiop:
@mparno I tried to improve the overload of the "+" operator in matlab.

Before the last commits `mset+mset2` in matlab was the only equivalent of C++ `mset += mset2`
Now in addition to the previous syntax you can do `mset=mset+mset2` (which is a more natural for matlab).
The issue is that, if you do `mset3=mset+mset2` (using another output variable name), now `mset3` and `mset` are the same object and modifying one will modify the other.

I am not sure how bad this is an issue or not.

mparno:
@rubiop I made some changes to the `MultiIndexSet` class that allows us to call the `+` operator on the sets themselves.  I updated the matlab binding so that calling `+` makes a copy, which I think is safer.   Can you check to make sure this still works on your end?  My matlab license ran out...

rubiop:
@mparno thanks Matt, `+=` in Matlab doesn't really exist so it is way better like this. I fixed a small typo and things worked as attended

mparno (approved):

PR #223, opened 2022-09-02
Linearized basis

### LINKED ISSUE #59: Add linearized-tail parameterization
- [x] Enable constructing basis function outside MonotoneComponent
- [x] Add linearized-tail 1d parameterization
- [x] ~Create new MapFactory method that takes in samples or quantiles and constructs map component with either linearized-tail or other parameterization~

mparno:
Closes #59 

- Added LinearizedBasis function that is templated on another 1d basis function type and linearizes the other basis outside some interval (lb,ub)
- Added the linearization interval as an option in the map options class.
- Included the linearization options in the MapFactory functions.
- Added linearization options to language bindings.


rubiop (commented):
I don't know if it is a Matlab only problem but if I am using the default option I have no linearization (which is expected considering the default values) and when I am specifying a number linearization occurs at -3 and 3 with whatever value I set.

I think would not see this issue since it is also using the default -3,3 values

mparno:
Good catch @rubiop !   That was indeed a bug on the c++ side.  I added a test to catch it and also fixed it.  Let me know if you run into any issues.

rubiop (approved):
Works, thanks @mparno !

PR #224, opened 2022-09-06
Poly normalization

mparno:
- Added ability to normalize orthogonal polynomials so that the inner product of p_n(x) with p_n(x) is 1.0 (with the weighting corresponding to the polynomial family).
- Added `basisNorm` option to `MapOptions` to allow normalization to be controlled during map construction.
- Added `basisNorm` option to all language bindings.


rubiop (approved):
works great!

PR #226, opened 2022-09-07
Functionality for computing off-diagonal derivatives

### LINKED ISSUE #17: Add full grad_x to monotone component
- Add full gradient computation to allow derivatives of conditional maps

mparno:
Closes #17 

- Added `Gradient` function to `ParameterizedFunctionBase` and all its children.
- Enables computing the gradient (or action of the transposed Jacobian matrix) of the function/map output with respect to the input.  
- Added language bindings.

Below is a python example of how to use this new functionality to plot the derivatives of one two-dimensional map component.  The resulting plot is also shown below. 

@michael-c-brennan This should allow you to propagate gradients through your composed maps, but let me know if don't think this will work for you.

@rubiop I added the matlab bindings, but it would be great if you could check to see if they work for you as well.

@dannys4 Same for the Julia bindings.  Could you check to see if you can use the new `Gradient` function?

Derivatives

```python
import numpy as np
import mpart as mt 
import matplotlib.pyplot as plt 
import matplotlib as mpl

dim = 2
totalOrder = 5
mset = mt.MultiIndexSet.CreateTotalOrder(dim, totalOrder).fix(True)

opts = mt.MapOptions()
comp = mt.CreateComponent(mset, opts)

coeffs = np.random.randn(comp.numCoeffs)
comp.SetCoeffs(coeffs)

nplot = 100
X1, X2 = np.meshgrid(np.linspace(-3,3,nplot),np.linspace(-3,3,nplot))
x = np.zeros((dim,nplot**2))
x[0,:] = X1.ravel()
x[1,:] = X2.ravel()

T = comp.Evaluate(x).reshape(nplot,nplot)
grad = comp.Gradient(x,np.ones((1,nplot**2)))

fig, axs = plt.subplots(ncols=3,sharey=True)
h = axs[0].pcolormesh(X1,X2,T)
fig.colorbar(h, ax=axs[0])
axs[0].set_title("Random $T_2(x_1,x_2)$")

norm = mpl.colors.SymLogNorm(linthresh=10, vmin=-5000.0, vmax=5000.0, base=10)
h = axs[1].pcolormesh(X1,X2,grad[0,:].reshape(nplot,nplot), norm=norm, cmap='coolwarm')
fig.colorbar(h, ax=axs[1])
axs[1].set_title("$\partial T_2 / \partial x_1$")


h = axs[2].pcolormesh(X1,X2,grad[1,:].reshape(nplot,nplot), norm=norm, cmap='coolwarm')
fig.colorbar(h, ax=axs[2])
axs[2].set_title("$\partial T_2 / \partial x_2$")

for ax in axs:
    ax.set_xlabel('$x_1$')
axs[0].set_ylabel('$x_2$')

plt.show()

```


michael-c-brennan (approved):
Looks great and everything seems to work for me!

Issue #231, opened 2022-09-12
NVCC Compilation Stalls on `MapFactory.cpp`


mparno:
The MapFactory class instantiates many different template options and results in VERY long compilation times, especially when compiling with CUDA and NVCC.   Breaking MapFactory.cpp into smaller pieces or explicit instantiated some of the classes in otehr cpp files might help us get around this.

dannys4:
Now that we merged #239 , I think we can close this? I might open up another (lower-priority) issue to refactor the files to make them more readable-- it's hard to tell which file contains what.

The things I would do for partial template instantiation using integers/enums ☹️ .

Issue #115, opened 2022-07-08
Special classes for common simple transformations


michael-c-brennan:
- [ ] `Scale` class: Apply matrix $x \mapsto Ax$, where $A$ is an invertible matrix
- [ ] `Rotate` class: Apply unitary matrix $x \mapsto Ux $, where $U$ is a unitary matrix
- [ ] `Shift` class: Apply shift $\ x \mapsto x + \mu $ where $\mu$ is vector

PR #239, opened 2022-09-20
 Linear algebra tools as well as `AffineMap` and `AffineFunction` classes

mparno:
- Added some wrappers around Eigen/CUBLAS for matrix multiplication and LU decompositions of matrices stored in Kokkos views.
- Added a new child of `ConditionalMapBase` called `AffineMap` that computes invertible maps of the form $Ax + b$.  If the matrix $A$ is rectangular, only the right square block of the matrix needs to be invertible.
- Added a new child of `ParameterizedFunctionBase` called `AffineFunction` that computes general (possibly non-invertible) functions $Ax+b$.
- Added detailed docs and code snippets showing usage of linear algebra tools.
- Fixed a bug in Array conversions when converting a const strided Kokkos view to Eigen.


mparno:
Closes #231 
Tackles part of #115

michael-c-brennan (approved):
Looks great to me! I think this will really helpful for future development.

PR #244, opened 2022-09-22
Add Kokkos array conversions

dannys4:
To add GPU support for Julia (and, presumably, MATLAB), we need the `ToKokkos` functions working with DeviceSpace.

Additionally, this fixes a few small bugs with tests in the `TestArrayConversions`.

### DIFF
--- a/MParT/Utilities/ArrayConversions.h
+++ b/MParT/Utilities/ArrayConversions.h
@@ -42,16 +42,16 @@ namespace mpart{
         @return A Kokkos view wrapping around the memory pointed to by ptr.
         @tparam ScalarType The scalar type, typically double, int, or unsigned int.
     */
-    template<typename ScalarType>
-    inline Kokkos::View<ScalarType*,Kokkos::HostSpace> ToKokkos(ScalarType* ptr, unsigned int dim)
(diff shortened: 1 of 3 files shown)


dannys4:
Be careful about the tests-- for some reason they fail on my local machine just because random numbers get chosen weirdly, so it's just barely not within epsilon. Perhaps test on your machine to check this

mparno:
Does Julia have arrays that live in GPU memory that you're planning on wrapping in a Kokkos::View?   Otherwise I'm not sure how this would be used.  It doesn't hurt to add it either way, but I'm curious what you have in mind.

dannys4:
This adds functions `ToKokkos` so that I can put a pointer from the host and it returns a Kokkos::View on the device. The objective is to make something similar to the python bindings, where they automatically copy over host to device if the map is on device. Since Julia uses `ToKokkos` for most of its array conversions, I'm just creating something similar to what we did with Eigen, where if `ToKokkos<...,DeviceSpace>` is called, it'll move the pointer (assumed to be on host) to device with no fuss, so we can just use practically the same function wrappers for the device-based map functions/classes.

mparno:
I see.  So this is essentially combining the `ToKokkos` and `ToDevice` functions.  Got it.

mparno (approved):

PR #246, opened 2022-09-23
Adding bindings for Julia affine and composed map classes

dannys4:
Commented out `EvaluateUntilK` due to odd `Layout` requirements.

### DIFF
--- a/bindings/julia/CMakeLists.txt
+++ b/bindings/julia/CMakeLists.txt
@@ -8,6 +8,7 @@ set(JULIA_BINDING_SOURCES
   src/ParameterizedFunctionBase.cpp
   src/TriangularMap.cpp
   src/Wrapper.cpp
+  src/AffineMap.cpp
   ../common/src/CommonUtilities.cpp
 )
 
--- a/bindings/julia/CMakeLists.txt
+++ b/bindings/julia/CMakeLists.txt
@@ -9,6 +9,7 @@ set(JULIA_BINDING_SOURCES
   src/TriangularMap.cpp
   src/Wrapper.cpp
   src/AffineMap.cpp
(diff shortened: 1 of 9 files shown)


dannys4:
@rubiop can you try this out and make sure it compiles with Julia on your machine? Just want to get this on the grid at some point soon

dannys4:
I'm currently encountering an issue with `ComposedMap`:
```julia
using MParT
A = randn(5,8)
B = randn(5,5)
pts = randn(8,2)
T1 = AffineMap(A)
T2 = AffineMap(B)
cmap = ComposedMap([T1,T2])
```

If you do `Evaluate(T1,pts)`, you get the right result, and `Evaluate(T2,A*pts)`, you get the right result, but for some reason, performing `Evaluate(cmap, pts)` throws a major error and crashes everything. Digging around a little bit, somehow it's performing a dGEMM $$C = A*B+C$$ with $$A\in\mathbb{R}^{5\times 8}$$, $$B\in\mathbb{R}^{8\times 2}$$ and $$C \in\mathbb{R}^{8\times 2}$$. I haven't found out why $$C$$ corresponds to that dimensional matrix, but I think it's safe to guess what the rest are according to the dimensions.

dannys4:
Ah, `ComposedMap` isn't supposed to work with rectangles. My bad.

dannys4:
looks like `jlmalloc` as written in the `JuliaArrayConversions` can sometimes allocate `NaN` values and, since we perform `C=AX+0C`, and IEEE states that `0*NaN=NaN`, we cannot recover from the original `NaN` value.

rubiop (approved):
tests succesful on my machine!

Issue #106, opened 2022-07-05
Add slice and condition functions to ConditionalMapBase


mparno:
In the `ConditionalMapBase` hierarchy, we should add functions that make it trivial to extract components of the map.  We can mimic the functions in eigen and provide:
- `head(unsigned int n)` Given a map T_{1:N}(x_{1:M}), this function will return the first n components: T_{1:n}(x_{1:M})
- `tail(unsigned int n)` Returns T_{N-n:N}(x_{1:M})
- `segment(unsigned int startInd, unsigned int n)` Returns T_{i:i+n}(x_{1:M}) 

In addition to these slice functions, we can add a function to define a conditional map by fixing the first n inputs:
- `given(Kokkos::View<double*> const& xstart)` returns a map T_{1:N}(x_{n:M}) where the first x_{1:n} block of the input is fixed to the input argument xstart.    Note that this function could be implemented in at least two ways: (1) We can keep the original map object the same, but add another class that concatenates `xstart` with a new values of `x_{n:M}` before calling the original map or (2) we could define the conditional map by updating the coefficients of the original map to account for the known value of `xstart`.   Option (1) is more general and will work for any implementation of `ConditionalMapBase` while option (2) will result in more efficient map evaluations, but requires a more intrusive implementation that may be restricted to the `MonotoneComponent` class.

PR #251, opened 2022-09-30
Add slicing for `TriangularMap` class

dannys4:
This PR adds the ability to slice a `TriangularMap` (using a contiguous slice in Julia/Python). I added bindings for those two languages (sorry @rubiop, no matlab ), but I haven't tested Python yet.

I had to change the return type of `CreateTriangular` for this to actually work in practice, since you can't slice arbitrary objects of type `ConditionalMapBase` (or use `GetComponent`). If this presents a problem, or we used `ConditionalMapBase` for a particular purpose, then I'd be happy to figure out what the best way forward is. Tagging @mparno particularly for this reason.

### DIFF
--- a/MParT/MapFactory.h
+++ b/MParT/MapFactory.h
@@ -4,6 +4,7 @@
 #include "MParT/MapOptions.h"
 
 #include "MParT/ConditionalMapBase.h"
+#include "MParT/TriangularMap.h"
 #include "MParT/MultiIndices/FixedMultiIndexSet.h"
 
 #include <math.h>
@@ -34,7 +35,7 @@ namespace mpart{
     auto map = MapFactory::CreateTriangular<Kokkos::HostSpace>(inDim, outDim, totalOrder, options);
 
      @endcode
-     
+
      */
     namespace MapFactory{
 
@@ -61,7 +62,7 @@ namespace mpart{
             @param options Additional options that will be passed on to CreateComponent to construct each MonotoneComponent.
          */
         template<typename MemorySpace>
-        std::shared_ptr<ConditionalMapBase<MemorySpace>> CreateTriangular(unsigned int inputDim,
+        std::shared_ptr<TriangularMap<MemorySpace>> CreateTriangular(unsigned int inputDim,
                                                                           unsigned int outputDim,
                                                                           unsigned int totalOrder,
                                                                           MapOptions options = MapOptions());
@@ -79,7 +80,7 @@ namespace mpart{
 
 
 
-        /** This struct is used to map the options to functions that can create a map component with types corresponding 
+        /** This struct is used to map the options to functions that can create a map component with types corresponding
             to the options.
         */
         template<typename MemorySpace>
@@ -98,14 +99,14 @@ namespace mpart{
                 auto iter = factoryMap->find(optionsKey);
                 if(iter == factoryMap->end())
                     throw std::runtime_error("Could not find registered factory method for given MapOptions.");
-                
+
                 return iter->second;
             }
 
             static std::shared_ptr<FactoryMapType> GetFactoryMap()
             {
                 static std::shared_ptr<FactoryMapType> map;
-                if( !map ) 
+                if( !map )
                     map = std::make_shared<FactoryMapType>();
                 return map;
             }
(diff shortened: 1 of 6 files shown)


mparno (commented):
Hey @dannys4 !   I think returning the triangular map is fine, but I think we might need to revise how the slicing is being performed.

I think it might be best to implement the `Slice` function in the `ConditionalMapBase` class and to have it slice by dimension, not by the potentially vector-valued components that make up the TriangularMap.   I think this is more natural, although harder to implement, because it slices the map in a way that is equivalent to slicing the vector-valued output of the map.  (i.e., `map.Evaluate(x)[a:b] ==  map.Slice(a,b).Evaluate(x)`).   If the boundaries of the slice land in the middle of one of the component outputs, we can then slice that block, which would be possible if the `Slice` function was added to `ConditionalMapBase`    What do you think?

I'll leave it up to you, but I could imagine a few different options for the default `ConditionalMapBase::Slice` implementation.  We could leave it as pure virtual, we could throw a not implemented exception, or we could define and return  a `SlicedMap` class that simply evaluates the full map and slices the output.

Also for future reference, it's possible to downcast a shared pointer to a parent class (e.g., `ConditionalMapBase`) to a child class (e.g., `TriangularMap`) by using something like `auto childPtr = std::dynamic_pointer_cast<ChildClass>(parentPtr);`.

dannys4:
It occurred to me that ConditionalMapBase should probably have a slice function, and I quite like the equivalence of `map.Evaluate(x)[a:b] == map.Slice(a,b).Evaluate(x)`. What do you think of two functions, `Slice` and `BlockSlice`. What I implemented here becomes `BlockSlice`-- `map.BlockSlice(2,4)` is really taking the second and third "block" of the map (python notation). This means you just have to create a new vector with a few of the same shared pointers. This would probably stay as a purely `TriangularMap` function, at least unless you have a better idea.

`Slice` would be a little more work, but would be exactly what you recommend-- you just take the subsection of `ConditionalMapBase`s corresponding to the actual concrete outputs, then recursively call `Slice` on the ends (assuming they're "jagged" in a sense). This would be a virtual function, so then things like `AffineMap` could implement them easily.

mparno:
That sounds good to me.

dannys4:
For completeness (and linking a mention), this partly solves #106 , but leaves `Condition` on the table.

dannys4:
@mparno need a nugget of C++ wisdom. Would you expect `Slice` and `BlockSlice` to be decorated as `const` functions? They theoretically don't change the contents of the map, but since everything is a shared pointer, modifying a slice will sometimes modify the original map (I would say "most likely" modifies the map, but there are easy-to-qualify conditions). I don't want to mark them as `const` and be misleading here.

dannys4:
I have run into a moderately large problem-- for most cases, I end up calling `std::shared_ptr<ConditionalMapBase> MonotoneComponent::Slice(int a, int b)`. I was thoughtlessly just returning `std::shared_ptr<...>(this)`, which I realized presents a problem when `this` MonotoneComponent is in a different shared pointer that gets deleted beforehand. Currently, this throws phenomenon happens and segfaults the program when C++ tries double deleting the same `MonotoneComponent`. While I'm not sure about the solution, I believe creating `std::shared_ptr Slice(int,int)` and `ConditionalMapBase SliceImpl(int,int)` in `ConditionalMapBase` (where each child class would implement `SliceImpl` with an eventual default implementation in `ConditionalMapBase`, which just returns some kind of generic `SlicedMap`)

dannys4:
see: https://en.cppreference.com/w/cpp/memory/enable_shared_from_this

mparno:
@dannys4 Just curious, where does this stand?   I have some upcoming applications where this feature would be great to have.  I know things will probably be slow over the the next couple of holiday weeks, but I would be happy to help push this over the finish line.

dannys4:
To be completely frank, this is a little lower than I'd like on my priorities list. What's left, to my knowledge, is making the `MonotoneComponent` inherit from the `std::enable_shared_from_this`. Past that, I'm not exactly sure what's left-- that's just the pressing problem, but I'm hoping that's all we need to "push it out this over the finish line". I'm trying to get distributions out the door right now, but I can focus on slicing afterward.

Issue #259, opened 2022-10-14
MATLAB Dockerfile


dannys4:
This is more of a posting, perhaps prefacing official support for a docker image for matlab bindings. This works on my computer, and I think should work on others.

Change your directory to `/your/path/MPART_DOCKER_MATLAB/` and make a file `Dockerfile` (no file extension)

```Dockerfile
# We pull from the MATLAB official image
FROM mathworks/matlab:latest

# Everything has to be done as root
USER root

# Install all the dependencies
RUN apt-get update && \
    apt-get install -y git cmake g++-9 libstdc++6

# Ensure we're using a new version of g++
RUN update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-9 100

# Clone and install Eigen 3.4
RUN git clone --depth 1 https://gitlab.com/libeigen/eigen.git --branch 3.4 && \
    cd eigen && \
    mkdir build && \
    cd build && \
    cmake .. && \
    make && \
    make install

# Clone and install Kokkos
RUN git clone --depth 1 https://github.com/kokkos/kokkos.git --branch 3.6.01 && \
    mkdir kokkos/build && \
    cd kokkos/build && \
    cmake \
        -DKokkos_ENABLE_SERIAL=ON \
        -DKokkos_ENABLE_OPENMP=ON \
        -DKokkos_ENABLE_CUDA=OFF \
        -DKokkos_ENABLE_HIP=OFF \
        -DKokkos_ENABLE_SYCL=OFF \
        -DBUILD_SHARED_LIBS=ON    \
        -DKokkos_CXX_STANDARD=17  \
        ../ && \
    make -j$(nproc) && \
    make install && \
    cd / && \
    rm -rf kokkos

# Set default user as matlab
ENV M_USER=matlab
ENV MPART_HOME=/home/${M_USER}/MParT

# Set up and compile Matlab
RUN git clone --depth 1 https://github.com/MeasureTransport/MParT.git --branch dannys4/matlab-compile && \
    cd MParT && \
    mkdir build && \
    cd build && \
    cmake -DMPART_BUILD_TESTS=OFF \
         -DMPART_PYTHON=OFF \
         -DMPART_JULIA=OFF \
         -DMPART_FETCH_DEPS=OFF \
         -DMPART_MATLAB=ON \
         -DCMAKE_INSTALL_PREFIX=${MPART_HOME} \
         ../ && \
    make && \
    make -j$(nproc) install && \
    cd / && \
[...]

PR #260, opened 2022-10-14
Adapt MexPlus to enable interleaved complex

dannys4:
Long story short, Mathworks changed how complex numbers were represented in 2018. While we don't really use complex numbers, it was actually easier to adapt Mexplus to support the new complex numbers than it was to remove them from the library. The old way is deprecated but not removed (though it will be at some point). CMAKE defaults to compiling MEX with the old standard, so we haven't had an issue, but if you want to test this for both, change `bindings/matlab/CMakeLists:18` between

` matlab_add_mex(NAME MParT_ SRC ${MEX_SOURCE} LINK_TO mpart Kokkos::kokkos Eigen3::Eigen R2017b)`

and

` matlab_add_mex(NAME MParT_ SRC ${MEX_SOURCE} LINK_TO mpart Kokkos::kokkos Eigen3::Eigen R2018a)`

if they both compile correctly and you can run some simple files, then everything should be A-okay.

### DIFF
--- a/bindings/matlab/CMakeLists.txt
+++ b/bindings/matlab/CMakeLists.txt
@@ -15,7 +15,7 @@ set(MEX_SOURCE
     src/MexMapOptionsConversions.cpp
 )
 
-matlab_add_mex(NAME MParT_ SRC ${MEX_SOURCE} LINK_TO mpart Kokkos::kokkos Eigen3::Eigen)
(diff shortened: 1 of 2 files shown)


dannys4:
For reference, this is required for #259 , since the MATLAB docker image seems to only support the newer standard for one reason or another

rubiop (approved):
Works on mac os and ubuntu

PR #262, opened 2022-10-17
261 triangularmaps constructor bug

michael-c-brennan:
1. Changed TriangularMap constructor to allow for map components of different dimensions.
2. Added a test for a TriangularMap made with 3 smaller TriangularMaps of different dimensions.


rubiop (approved):
Looks great!

PR #270, opened 2022-10-31
SummarizedMap, SingleEntryMap, improved Python tests

michael-c-brennan:
`SummarizedMap`: a class for maps of the form $T(x) = T(s(x_{< n }), x_n)$ where $s: \mathbb{R}^{n-1} \to \mathbb{R}^r$ is a _summary function_, that summarizes the leading input of the map. We assume $r < n-1$.

`MapFactory::CreateSingleEntryMap`: Factory method to create maps that are the identity for all but one entry.

Improved Python tests: 
- Small bug fix in `test_ConditionalMapBase.py`
- Uncommented quadrature tests that were commented


MParT/MapFactory.h
```diff
@@ -4,6 +4,8 @@
 #include "MParT/MapOptions.h"
 
 #include "MParT/ConditionalMapBase.h"
+#include "MParT/SummarizedMap.h"
+//#include "MParT/DebugMap.h"
```
dannys4, line 8: We'll need to make an issue introducing a "debug build option" of MParT so this will auto build when we make a debug build, but won't otherwise. I'll put it up as an issue after this.

dannys4 (changes requested):
Please address these

michael-c-brennan:
Thanks @dannys4! I believe I addressed all of them, but let me know if I missed anything.

PR #272, opened 2022-11-10
Maintaining GPU Code

dannys4:
There are some problems that, while implementing the Distribution class, I found. I started fixing them in my branch for distributions, but it slowly became apparent that there were some quirks that kept coming up which needed attention. This ameliorates some of the issues I was facing. Namely,
- Changes `SetCoeffs` to make more sense. Now, subclasses should only need to implement `SetCoeffs` when the new coeffs are in the same `MemorySpace` as the map. Otherwise, PFB should take care of moving things between host and device.
- Changes `WrapCoeffs` to make more sense. You should only be able to wrap coefficients when they live in the same `MemorySpace` as the map-- it doesn't make sense/shouldn't be within scope to handle keeping a host + device copy synchronized since that would slow down the functions considerably.
- Adds Cholesky factorization
- Adds simple tests for Cholesky and LU factorizations
- Replace `#if defined(KOKKOS_ENABLE_CUDA) || defined(KOKKOS_ENABLE_SYCL)` with `#if defined(MPART_ENABLE_GPU)`.


dannys4:
Currently, building this on GPU fails big time. I'm working on why

dannys4:
Closing this in favor of #386

PR #276, opened 2022-11-15
275 bug in affine function

### LINKED ISSUE #275: Bug in `AffineFunction` with matrix `A` going out of scope
I have the following script for creating a ComposedMap out of many SingleEntryMaps, where each SingleEntryMap uses a SummarizedMap component:
```
maps = []

# first components don't need summaries
for idx in range(0,lrc_rank+1):
	mset = mt.MultiIndexSet.CreateTotalOrder(idx+1, total_order, noneLim)
	component = mt.CreateComponent(mset.fix(True)  , map_options)
	maps.append(mt.CreateSingleEntryMap(dim, idx+1, component))


summary_matrices = []
for idx in range(lrc_rank+1, dim):

	# create summary_matrix
	A = np.random.randn(idx, idx)
	Q, _ = np.linalg.qr(A)
	summary_matrix = Q[:lrc_rank,:].copy()

	summary_matrices.append(summary_matrix) # if you uncomment this, things work for lrc_rank = 1
	summary_fun = mt.AffineFunction(np.asfortranarray(summary_matrix))
	mset = mt.MultiIndexSet.CreateTotalOrder(lrc_rank+1, total_order, noneLim)
	component = mt.CreateComponent(mset.fix(True), map_options)
	summarized_component = mt.SummarizedMap(summary_fun, component)
	single_entry_map = mt.CreateSingleEntryMap(dim, idx+1, summarized_component)
	maps.append(single_entry_map)

transport_map = mt.ComposedMap(maps, dim)
coeffs = np.random.rand(transport_map.numCoeffs)
coeffs = coeffs/(np.linalg.norm(coeffs))
transport_map.SetCoeffs(coeffs)
```
Here `lrc_rank = 1 or 2`, small enough that each map component is cheap to evaluate. When I try to evaluate the map for a large-ish dimension (e.g. 100), I would get an error: `ERROR: In MonotoneIntegrand, A nan was encountered in value of g(df(...))`. Today @dannys4, @rubiop and I found that the values held inside `A_` of the AffineFunction object held nonsense values (with scales like 1E-160 or 1E30) using gdb. This caused the summary computed by SummarizedMap to be nonsense, and lead to the nonsense evaluation. We think this is because the variable `summary_matrix` used to create each AffineFunction gets collected by the python garbage collector after creating the AffineFunction (although disabling automatic garbage collection using `gc.disable()` doesn't help). We tried to hack it so that each `summary_matrix` would stay in scope by storing them in `summary_matrices`, which worked for `lrc_rank = 1` but it still breaks for `lrc_rank = 2` (I have no idea why). I'm looking for ideas for how to fix this.

michael-c-brennan:
- Deep copying of `A` and/or `b` in constructors for AffineMap and AffineFunction
- Added python unit tests of AffineMap and AffineFunction

Fixes #275

### DIFF
--- a/MParT/AffineFunction.h
+++ b/MParT/AffineFunction.h
@@ -46,8 +46,6 @@ public:
                       StridedMatrix<double, MemorySpace>              output) override;
 
 
-    void print_Aij(int i, int j) { std::cout << "A(" << i << "," << j << ") = " << A_(i,j) << std::endl; }
-    void print_ptr_A() { std::cout << "ptr A = " << A_.data() << std::endl; }
 
 
 protected:
--- a/MParT/AffineFunction.h
+++ b/MParT/AffineFunction.h
@@ -45,6 +45,11 @@ public:
                       StridedMatrix<const double, MemorySpace> const& sens,
                       StridedMatrix<double, MemorySpace>              output) override;
 
+
+    void print_Aij(int i, int j) { std::cout << "A(" << i << "," << j << ") = " << A_(i,j) << std::endl; }
+    void print_ptr_A() { std::cout << "ptr A = " << A_.data() << std::endl; }
+
+
 protected:
 
     StridedMatrix<double,MemorySpace> A_;
--- a/MParT/AffineFunction.h
+++ b/MParT/AffineFunction.h
@@ -46,6 +46,8 @@ public:
                       StridedMatrix<double, MemorySpace>              output) override;
 
 
+    // void print_Aij(int i, int j) { std::cout << "A(" << i << "," << j << ") = " << A_(i,j) << std::endl; }
+    // void print_ptr_A() { std::cout << "ptr A = " << A_.data() << std::endl; }
 
 
 protected:
--- a/bindings/python/tests/test_AffineFunction.py
+++ b/bindings/python/tests/test_AffineFunction.py
@@ -1,0 +1,98 @@
+# MultiIndex test
+
+import mpart
+import numpy as np
+
+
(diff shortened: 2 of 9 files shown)


MParT/AffineFunction.h
```diff
@@ -45,6 +45,11 @@ class AffineFunction : public ParameterizedFunctionBase<MemorySpace>
                       StridedMatrix<const double, MemorySpace> const& sens,
                       StridedMatrix<double, MemorySpace>              output) override;
 
+
+    void print_Aij(int i, int j) { std::cout << "A(" << i << "," << j << ") = " << A_(i,j) << std::endl; }
+    void print_ptr_A() { std::cout << "ptr A = " << A_.data() << std::endl; }
```
dannys4, line 50: Comment these out for now? We'll tie them to the `MPART_DEBUG` issue.

bindings/python/tests/test_AffineFunction.py
```diff
@@ -0,0 +1,98 @@
...
+import mpart
+import numpy as np
+
+
+def test_ShiftOnlyFunction():
+
+
+    b = np.array([1.,2.]).reshape(2,1)
+    shift = mpart.AffineFunction(b)
+    del b  # checking that AffineFunction deep-copied b; see issue 275
```
dannys4, line 12: Smart way of testing!

dannys4 (changes requested):
Check out the debugging stuff. Otherwise Looks good!

dannys4 (approved):

PR #280, opened 2022-11-16
279 triangularmap handing map components without coefficients like identitymap

michael-c-brennan:
This adds checks to the CoeffGrad and LogDeterminantCoeffGrad functions in `TriangularMap` so that components with no coefficients are skipped. It also adds a test for `SingleEntryMap` that presented the issue.


dannys4 (approved):
✅

PR #281, opened 2022-11-18
Save and load using cereal

dannys4:
Added compiler flag `MPART_ARCHIVE`, which turns on everything serialization-related. MParT will automatically (and minimally) install the `cereal` headers if it cannot find the `cereal` library. This adds serialization for
- `FixedMultiIndexSet<MemorySpace>`
- `Kokkos::View<ScalarType*(*), MemorySpace>`
- `MapOptions`

I also added a few functions for conveniently accessing the `FixedMultiIndexSet` inside of a `MonotoneComponent`. Currently there are (untested) bindings for Python to serialize and deserialize
- `FixedMultiIndexSet<MemorySpace>`
- `ParameterizedFunctionBase<MemorySpace>::Coeffs()`
- `MapOptions`
Note the middle will have a function that serializes a PFB, but when you deserialize, you just get an `Eigen::Ref` since it doesn't keep track of the map type when serializing. See the test for an example.


dannys4:
This lacks documentation and Julia/Matlab bindings.

MParT/Utilities/Serialization.h
```diff
@@ -0,0 +1,65 @@
...
+#include <cereal/types/base_class.hpp>
+#include <cereal/archives/binary.hpp>
+#include <Kokkos_Core.hpp>
+
+#include "MParT/Utilities/ArrayConversions.h"
+
+namespace cereal {
+    template <typename ScalarType, typename MemorySpace, typename Archive>
+    std::enable_if_t<traits::is_output_serializable<BinaryData<ScalarType>, Archive>::value, void> save(
+        Archive &ar, Kokkos::View<ScalarType*, MemorySpace> const &vec) {
```
mparno, line 13: Should this be for arbitrary memory spaces or should we restrict it to HostSpace only?   Instead of throwing an exception in the `load` function, we could just restrict ourselves to the hostspace.   

This will also only work for views with the default layout.  We often use strided layouts, especially when interacting with other languages like python, and should allow for saving with arbitrary layouts and then loading back into either `LayoutLeft` or `LayoutRight`.   I'm not sure it makes sense to load back into a strided view.

tests/Test_Serialization.cpp
```diff
@@ -0,0 +1,76 @@
...
+        MapOptions options2;
+        iarchive(options2);
+        REQUIRE(options1.basisType == options2.basisType);
+        REQUIRE(options1.basisNorm == options2.basisNorm);
+        unsigned int inputDim2, outputDim2, maxDegree2;
+        iarchive(inputDim2, outputDim2, maxDegree2);
+        REQUIRE(inputDim2 == triMap1.inputDim);
+        REQUIRE(outputDim2 == triMap1.outputDim);
+        REQUIRE(maxDegree2 == maxDegree);
+        Kokkos::View<double*,Kokkos::HostSpace> coeffs2;
```
mparno, line 59: Consider adding tests with other layouts.

mparno (changes requested):
Looks great overall, but I think we should add the ability to serialize Kokkos views with layouts other than the default.

mparno (approved):
Looks great!  Nice job Danny.

PR #290, opened 2022-12-09
Adding KeepCoeffs option to TriangularMap and ComposedMap

michael-c-brennan:
Currently, both TriangularMaps and ComposedMaps assume their components do not have their coefficients set (which is often the case). There are cases where we want to build maps after optimizing individual components. This PR adds the ability to tell the constructors to keep the coefficients of their components.


dannys4 (approved):
Added a few docs so people are aware of the feature.

PR #296, opened 2022-12-19
Add wavelets

dannys4:
TODO:
- [x] Add Ricker wavelet
- [ ] Add Ricker wavelet so we can easily use it in a map
- [ ] Test Ricker wavelet
- [ ] Integrate it into `MultivariateExpansion`
- [ ] Add option to choose between uniform and Normal quantiles
- [ ] Add class (similar to `LinearizedBasis`) to represent wavelets in just the last element and polynomials elsewhere (Push this off to new PR?)

### DIFF
--- a/MParT/MapOptions.h
+++ b/MParT/MapOptions.h
@@ -9,7 +9,8 @@ namespace mpart{
     {
         ProbabilistHermite,
         PhysicistHermite,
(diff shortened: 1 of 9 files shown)


dannys4:
keeping branch open, but closing the PR.

PR #297, opened 2022-12-20
Add Cholesky Factorization, LU tests

dannys4:
Since #272 is somewhat stalled, I'm pulling the Cholesky factorization out and making it its own PR.

### DIFF
--- a/MParT/Utilities/LinearAlgebra.h
+++ b/MParT/Utilities/LinearAlgebra.h
@@ -2,6 +2,8 @@
 #define MPART_LINEARALGEBRA_H
 
 #include <Kokkos_Core.hpp>
+#include <Eigen/LU>
+#include <Eigen/Cholesky>
 
 #include "MParT/Utilities/ArrayConversions.h"
 #include "MParT/Utilities/GPUtils.h"
--- a/MParT/Utilities/LinearAlgebra.h
+++ b/MParT/Utilities/LinearAlgebra.h
@@ -296,7 +296,7 @@ public:
 
     /** Computes \f$L^{-1}B\f$ and stores the results in B.
     */
-    void solveLInPlace(Kokkos::View<double**,Kokkos::LayoutLeft,MemorySpace> B);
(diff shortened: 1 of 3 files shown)


MParT/Utilities/LinearAlgebra.h
```diff
@@ -270,6 +272,53 @@ class PartialPivLU
...
+    /** Computes \f$A^{-1}x\f$ and stores the results in x.
+    */
+    void solveInPlace(Kokkos::View<double**,Kokkos::LayoutLeft,MemorySpace> x);
+
+    /** Returns a view containing \f$A^{-1}x\f$. */
+    Kokkos::View<double**,Kokkos::LayoutLeft,MemorySpace> solve(StridedMatrix<const double,MemorySpace> x);
+
+    /** Computes \f$L^{-1}B\f$ and stores the results in B.
+    */
+    void solveLInPlace(Kokkos::View<double**,Kokkos::LayoutLeft,MemorySpace> B);
```
mparno, line 299: The `L` in the middle of the function name is not immediately obvious when I'm scanning the code.   What do you think about moving this to the end of the function name, so that we'd have `solveInPlaceL`?  Because `multiplyL`, has the `L` at the end, this seems more consistent and readable to me.

src/Utilities/LinearAlgebra.cpp
```diff
@@ -255,7 +306,187 @@ double PartialPivLU<mpart::DeviceSpace>::determinant() const
...
+void Cholesky<mpart::DeviceSpace>::compute(Kokkos::View<const double**,Kokkos::LayoutLeft,mpart::DeviceSpace> A)
+{
+
+    ldA = A.stride_1();
+
+    int nrows = A.extent(0);
+    int ncols = A.extent(1);
+    assert(nrows==ncols);
+
+    // Resize the space for storing the LU factorization
```
mparno, line 319: Change comment from LU to Cholesky of LLT

mparno (commented):
Great work @dannys4 !   I have a few minor suggestions, but otherwise I think this is ready to go.

mparno (approved):

PR #298, opened 2022-12-20
Add Density functionality, Gaussian density

dannys4:
A few classes.
- `SampleGenerator`: a class that just generates samples. In the future, could wrap around some kind of PDE solver with random elements or something of the sort.
- `DensityBase`: a class representing a real-valued nonnegative differentiable function. Just needs to evaluate the log and the grad of the log
- `Distribution`: this one gets tricky. I'm not sure I went about the best way to perform this, and @mparno can correct me if this is a bad idea, but I went ahead and made it so that we can use two different classes, one for the `SampleGenerator` and one for the `DensityBase` part of it. Then, this doesn't actually copy any data, it just stores those shared pointers so that we minimize overhead. I think there's a way to do it while inheriting both `SampleGenerator` and `DensityBase`, but there are some nuances regarding keeping track of which `dim_` field you want to use etc.
- `GaussianSamplerDensity`: a class that is both a `SampleGenerator` and a `DensityBase` representing a Gaussian. Depending on how you construct it, it can assume isometry or not, as well as optionally have a shift for the mean. Implementing a pure diagonal covariance is in future work, but should be rather straightforward (I'll file an issue shortly).
- `GaussianDistribution`: This just wraps around `Distribution<MemorySpace, GaussianSamplerDensity<MemorySpace>, GaussianSamplerDensity<MemorySpace>>`, which is a bit of a mouthful.

Note the tests for `GaussianDistribution` reflect #299 , which is a pain in the neck but easy to work around for the time being. Sorry for such a large PR-- I should've probably split the Gaussian into its own PR.


dannys4:
I changed so that `Distribution` isn't templated according to the density and sampler types

dannys4:
`GaussianSamplerDensity::LogDensityImpl` actually has a major race condition in this PR, but it gets fixed in #304 .

michael-c-brennan (approved):
Everything looks good to me!

PR #304, opened 2023-01-17
Add map training

dannys4:
This PR adds map objective and map training!!
- `MapObjective` is an abstract class to allow maps to train with an arbitrary objective. I'm thinking about how to get this working with external functions (i.e. from python/julia), but that'll be a bit. It keeps track of training and testing datasets since those are, in some sense, inherent to the fitting problem.
- `KLMapObjective` is a subclass of `MapObjective` that implements the usual KL map objective, from an arbitrary distribution
- `TrainOptions` are options largely for optimization with NLOpt as well as desired verbosity.
- `TrainMap` allows you to, well, train a given map according to the map objective and training options.

I will add bindings in a separate PR, since `TrainOptions` will likely be a pain.

### DIFF
--- a/MParT/TrainMap.h
+++ b/MParT/TrainMap.h
@@ -17,9 +17,28 @@ struct TrainOptions {
     double opt_ftol_rel = 1e-3;
     double opt_ftol_abs = 1e-3;
     double opt_xtol_rel = 1e-4;
-    int opt_maxeval = 30;
-    double opt_maxtime = 100.;
+    double opt_xtol_abs = 1e-4;
+    int opt_maxeval = 1000;
+    double opt_maxtime = std::numeric_limits<double>::infinity();
     bool verbose = false;
+    /**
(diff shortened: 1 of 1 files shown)


mparno:
@dannys4 Is this still a draft?

dannys4:
I have to merge main into it, which is nontrivial. I can have it done in about 20 minutes or so

mparno (changes requested):
I got a little ahead of things by reviewing #305 before this one.   @dannys4 Could you look at the comments for the `MapObjective.h` and `TrainMap.h` in #305?   Those are probably best addressed here.   The comments about the bindings can stay in #305.   Thanks!

dannys4:
I found the requested changes in `TrainMap.h` agreeable, but I'm not as sure about the path forward for `MapObjective`-- feel free to continue the conversation here or there on those points.

mparno (approved):

PR #305, opened 2023-01-28
Bindings for `trainMap` and Serialization

dannys4:
This performs the bulk of the work such that you can serialize a map, fixed mset, and mapoptions in each binding language, adds some features under the hood (e.g. `ReduceDim`), and of course binds the `TrainMap` function to each language


bindings/matlab/mat/GaussianKLObjective.m
```diff
@@ -0,0 +1,38 @@
+classdef GaussianKLObjective < handle
```
mparno, line 1: Is there a reason this is restricted to Gaussian reference distributions and not arbitrary densities like the `KLObjective` class in c++?   We should just expose the Distribution stack to Matlab if the lack of those classes is preventing us from mimicking the `KLObjective` class here.

bindings/python/src/MapObjective.cpp
```diff
@@ -0,0 +1,52 @@
...
+        .def("TestError", &KLObjective<MemorySpace>::TestError)
+        .def("TrainError", &KLObjective<MemorySpace>::TrainError)
+    ;
+
+    py::class_<KLObjective<MemorySpace>, MapObjective<MemorySpace>, std::shared_ptr<KLObjective<MemorySpace>>>(m, tName.c_str())
+        .def(py::init( [](Eigen::Ref<Eigen::MatrixXd> &train){
+            StridedMatrix<double, MemorySpace> trainView = MatToKokkos<double, MemorySpace>(train);
+            Kokkos::View<double**,MemorySpace> storeTrain ("Training data store", trainView.extent(0), trainView.extent(1));
+            Kokkos::deep_copy(storeTrain, trainView);
+            std::shared_ptr<GaussianSamplerDensity<MemorySpace>> density = std::make_shared<GaussianSamplerDensity<MemorySpace>>(trainView.extent(0));
```
mparno, line 33: Like for the matlab bindings, I don't think we should restrict ourselves to a standard normal reference distribution in the bindings if the c++ code is more general.

MParT/TrainMap.h
```diff
@@ -0,0 +1,40 @@
...
+namespace mpart {
+struct TrainOptions {
+    std::string opt_alg = "LD_LBFGS";
+    double opt_stopval = -std::numeric_limits<double>::infinity();
+    double opt_ftol_rel = 1e-3;
+    double opt_ftol_abs = 1e-3;
+    double opt_xtol_rel = 1e-4;
+    double opt_xtol_abs = 1e-4;
+    int opt_maxeval = 30;
+    double opt_maxtime = 100.;
```
mparno, line 22: Do we want to set a maximum time?  I'd prefer if this was infinity (or practically so), so that the optimizer doesn't stop prematurely after a long run because I forgot to set this to something larger.

MParT/TrainMap.h
```diff
@@ -0,0 +1,40 @@
...
+
+namespace mpart {
+struct TrainOptions {
+    std::string opt_alg = "LD_LBFGS";
+    double opt_stopval = -std::numeric_limits<double>::infinity();
+    double opt_ftol_rel = 1e-3;
+    double opt_ftol_abs = 1e-3;
+    double opt_xtol_rel = 1e-4;
+    double opt_xtol_abs = 1e-4;
+    int opt_maxeval = 30;
```
mparno, line 21: 30 seems really small for a maxeval....  How about 1000?

mparno (changes requested):

MParT/MapObjective.h
```diff
@@ -0,0 +1,52 @@
...
+#define MPART_MAPOBJECTIVE_H
+
+#include "ConditionalMapBase.h"
+#include "Distributions/PullbackDensity.h"
+#include "Utilities/ArrayConversions.h"
+#include "Utilities/LinearAlgebra.h"
+
+namespace mpart {
+template<typename MemorySpace>
+class MapObjective {
```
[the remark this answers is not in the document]
dannys4: I thought about this a lot actually. On one hand, I kind of agree, but on the other hand this means that we have to create a new objective on every iteration of ATM (or replace the map), or any time you want to do something that trains different maps several times on the same dataset. You could theoretically just add a `SetMap` function, but that's basically just doing the same thing as I'm doing right now! At the end of the day, I think it all ends up evening out, and I feel like the syntax `TrainMap(map, objective, options)` makes a lot of sense-- I like the mathematical idea of saying the objective is a functional $F(T;d)$ where $T$ is the transport map and $d$ is the data, and separating out `objective` from `map` is a good representation of that.

MParT/MapObjective.h
```diff
@@ -0,0 +1,52 @@
...
+    StridedVector<double, MemorySpace> TrainCoeffGrad(std::shared_ptr<ConditionalMapBase<MemorySpace>> map) const;
+    void TrainCoeffGradImpl(std::shared_ptr<ConditionalMapBase<MemorySpace>> map, StridedVector<double, MemorySpace> grad) const;
+
+    virtual double ObjectivePlusCoeffGradImpl(StridedMatrix<const double, MemorySpace> data, StridedVector<double, MemorySpace> grad, std::shared_ptr<ConditionalMapBase<MemorySpace>> map) const = 0;
+    virtual double ObjectiveImpl(StridedMatrix<const double, MemorySpace> data, std::shared_ptr<ConditionalMapBase<MemorySpace>> map) const = 0;
+    virtual void CoeffGradImpl(StridedMatrix<const double, MemorySpace> data, StridedVector<double, MemorySpace> grad, std::shared_ptr<ConditionalMapBase<MemorySpace>> map) const = 0;
+};
+
+template<typename MemorySpace>
+class KLObjective: public MapObjective<MemorySpace> {
```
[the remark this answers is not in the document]
dannys4: How would you feel about keeping `KLObjective` (and adding docs that this estimates the MLE) and then if we do the pushforward density, we use `ReverseKLObjective`? I'm okay with changing it to `MLEObjective` if you want, but that'll take a little bit of adjustment. Also, I like the idea of describing the objective as a function of what it does, not what the map would be at the minimizer of the objective.

MParT/TrainMap.h
dannys4: I've been finding that 30 is usually plenty for my scenarios (which are admittedly a little low on dimension), but we can change it if you want. If it's any consolation, this returns an improved map after it finishes, so if you forget to increase this, you can just throw it back into `TrainMap` a second time.

bindings/matlab/mat/GaussianKLObjective.m
dannys4: Good question. I figured that, if 50% of maps trained are normalizing samples, it's certainly worth having something that lets you easily get from samples to trained map in a few lines. Can we make this as an issue for a future PR? This would require a larger restructuring of code, and I think would be more fitting to fix in all languages at once.

mparno (approved):

PR #307, opened 2023-01-30
Adaptive Transport Map

dannys4:
This adds a few things. Mainly, `ATMOptions` and `mpart::AdaptiveTransportMap`.
- `ATMOptions` descends from both `MapOptions` and `TrainOptions` so it'll store everything you want to tweak within each map that ATM creates as well as everything you'll want to tweak for optimizing each map. It also stores a few custom ATM options: `maxSize`, which is the maximum number of coefficients that your map can have, `maxPatience`, which is the maximum number of iterations you can go in ATM without improving on your Test accuracy (i.e. accuracy on your test dataset), and `maxDegrees` which will limit the complexity of each input (e.g. `{25,4}` means that $x_1$ can only appear with a degree strictly less than 25 _anywhere_ in the map, and $x_2$ can only appear with a degree strictly less than 4 _anywhere_ in the map).
- `AdaptiveTransportMap` is exactly what it sounds like. It's a pretty simple function call that hides a tremendous amount under the hood. I'm not entirely sure it's implemented correctly, but I'm happy enough with the performance on a few test cases that I'm willing to push it out onto the world.
- `MultiIndex::AnyBounded` and `MultiIndexSet::FilterBounded` power the `maxDegrees` option for ATM
- Change default optimization algorithm to SLSQP, as LBFGS is a little bugged for ATM (I believe it's when you initialize close to the optimum, see [NLopt.jl/33](https://www.github.com/JuliaOpt/NLopt.jl/issues/33) for a related issue).
- Changing some of the CI pipeline apparently?
- Change verbosity to add a second level specifically for debugging.

I'm going to keep putting up pull requests until someone stops me.


mparno (approved):
Thanks Danny.  I'm sorry for my incredibly slow response on this review.   Overall everything looks great!  There are a few places where I think we could use `Limiter` feature of `MultiIndexSet` instead of manually checking for bounded multiindices, but that's something we just tweak later on.  

The only thing I think we should change, or at least discuss, is the the use of cerr and cout.  It seems like we're using cerr in a bunch of places that aren't throwing errors, but maybe I'm just misinterpreting your intent.   Once we address that, I'm happy to merge.

PR #310, opened 2023-02-03
Bindings for ATM

dannys4:
Bind necessary features to use ATM. Unfortunately, I could not _quite_ figure out a way to make everything exactly the same in all the languages, with two major dichotomies
- In C++, Julia, and Python, one uses `CreateGaussianKLObjective` to get a KLObjective with a gaussian "target" density (see below). For Matlab I sort-of had to make a `GaussianKLObjective` class unfortunately, so you instantiate it with `GaussianKLObjective` in those bindings.
- In C++, Julia, and Matlab, you pass in your initial `MultiIndexSet`s as a vector and the algorithm with change those sets in-place to be the optimal `MultiIndexSet`s (it returns a trained map using those msets). This is not easily doable in Python bindings that I found so, as a workaround, it just returns the msets as the first output and the trained map as the second output

However, this should work nicely, with a bonus feature of `ATMOptions::MaxDegrees` working in all the languages. I tidied some code up in the bindings as well, particularly regarding the Matlab bindings.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 build
+*.egg-info*
 .vscode
 docs/_build
 bin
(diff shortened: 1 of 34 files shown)


dannys4:
That looks like a smart solution, thanks. There were some merge conflicts from main, so I'll merge once the tests pass.

PR #311, opened 2023-02-09
Made MParT pip installable

mparno:


dannys4 (approved):
Looks reasonable enough. Sorry about forgetting to look for NLopt/cereal quietly.

.github/workflows/pypi-deploy.yml
```diff
@@ -0,0 +1,61 @@
+name: Build and upload to PyPI
+
+# Build on every branch push, tag push, and pull request change:
+on: [push, pull_request]
```
dannys4, line 4: Wait, are we sure we want to push to pypi on every single PR? is there any downside to this? I'm not exactly sure how versioning works for this stuff

dannys4: Nevermind, I see now it _builds_ for pypi at each PR, but only pushes to pypi at each tagged release. My bad.

dannys4 (approved):

PR #314, opened 2023-02-10
Fixing explicit instantiation must occur in namespace error.

mparno:
When trying to build with pip, I ran into this error:
```
GaussianSamplerDensity.cpp:134:17: error: explicit instantiation of 'mpart::GaussianSamplerDensity' must occur in namespace 'mpart'
```

This pull request should fix that.


dannys4 (approved):

dannys4:
I ran into this before and fixed it, but merging these PRs have been weird

PR #328, opened 2023-04-17
Mparno/issue327

### LINKED ISSUE #327: Affine map in python provides nans
At least the `.Evaluate()` method for `AffineMap` defined in python is buggy providing nans at random computation time. In my experience it doesn't depend on the value of the coefficients nor the points where the map is evaluated. For the same example in matlab I have no issue so there is probably something weird happening with pybind.

In this example:

```
import numpy as np
import os
os.environ['KOKKOS_NUM_THREADS'] = '8'
import mpart as mt
print('Kokkos is using', mt.Concurrency(), 'threads')
np.random.seed(1)
samples = np.random.randn(2,200000)
A = np.array([[ 3.97678524, 0.        ],
       [ 4.55559914,  0.01547669]])
c = np.array([ -7.94509863, -18.88155037])
L = mt.AffineMap(np.asfortranarray(A),np.asfortranarray(c))
samples_st = L.Evaluate(samples)
print(np.mean(samples_st))
```

The print function at the end will return nan if there is any nan returned by the `AffineMap`.

mparno:
Closes #327 

The bug in #327 stemmed from not initializing the output matrix to 0.0, which allowed nans to trickle through.  I'm not entirely sure why this wasn't popping up anywhere other than the python bindings.

- Also added support for passing in numpy arrays with arbitrary storage orders to AffineMap and AffineFunction.  This means we won't need to use the `asfortranarray` function anymore.
- Since we were making deep copies of the arrays in AffineMap anyway, I forced them to be LayoutLeft, so no check or copy is needed during the factorization.

### DIFF
--- a/MParT/AffineFunction.h
+++ b/MParT/AffineFunction.h
@@ -51,9 +51,9 @@ public:
 
 
 protected:
-
-    StridedMatrix<double,MemorySpace> A_;
-    StridedVector<double,MemorySpace> b_;
+    
+    Kokkos::View<double**, Kokkos::LayoutLeft, MemorySpace> A_;
+    Kokkos::View<double*, Kokkos::LayoutLeft, MemorySpace> b_;
 
     int ldA;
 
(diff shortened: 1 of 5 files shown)


dannys4:
That's crazy, I should've thought of this. I had a similar issue with the gaussian distributions a few months ago. This might fix the ARM problem, but we'll have to check it out

rubiop (approved):
Thanks for fixing so quickly!

PR #339, opened 2023-07-23
Remove asserts

dannys4:
Unfortunately, asserts kill Julia, so I tried to revise the code to adhere a little more strongly to the idea of "exceptions are for things that shouldn't happen; asserts are for things that can't happen".

I understand that the original reason we (matt) threw asserts around is for performance portability-- you cannot throw an exception in CUDA code-- so I instead used a templated error class with static function that defaults to asserting 0 and, if we're in host space, we throw an error. Please let me know if you'd prefer it different.


mparno (approved):
This looks good.  Thanks @dannys4!   I only have one suggestion for the future.  Now that we're not using asserts, we can provide some more information in the error messages.  If possible to include, I find it much easier to debug when the exception messages have addition information like the argument dimensions.  For example, instead of just saying "Invalid dimension of inputs."  We could say something like "Input dimensions don't match.  Input x has size 10 and input y has size 11."    This is already a big improvement, so I'm happy to merge, but something to consider in the future.

dannys4:
Fixing this now then I'll merge after tests pass

PR #346, opened 2023-09-21
Add sigmoid, move root finding

dannys4:
- Creates linear sigmoid 1d basis (often called "RBFs", we actually care about using the integral of the RBF, which is a sigmoid function in most literature)
- Allows for centers and widths to allow for more expressible functions
- Moves the 1d root finding to external file, which will be needed for multiple different function representations (and also cleans up the file)


mparno (approved):
Looks good.  Thanks @dannys4 !

PR #349, opened 2023-10-04
Fixed bug in `FixedMultiindexSet` python bindings

mparno:
Segfaults were occurring when manually creating a `FixedMultiIndexSet` in python because of a shallow copy of a temporary vector in the python bindings for the `FixedMultiIndexSet` constructor.  This PR fixes the issue with deep copies.

Also fixes a bug in the `ConstVecToKokkos` function that showed up when exploring fixes to the python bindings.


dannys4 (approved):
Thanks for updating all the versions. Looks good overall

PR #353, opened 2023-10-20
Integration Nugget

mparno:
Sometimes, maps can be more robust by forcing them to have a nonzero minimum slope.   This PR enables this by adding a nugget option to the Monotone Integrand class.  The new form of map component $T_d(x_{1:d})$ now has an additional fixed parameter $\epsilon\geq 0$ and takes the form
 
$$
T_d(x_d; x_{1:d-1}) = f(x_{1:d-1},0) + \int_0^{x_d}  \quad g(\partial_d f(x_{1:d},t)) + \epsilon  \quad dt
$$

The value of $\epsilon$ is set in `MonotoneComponent` construct but is also available in `MapOptions`.  All of the language bindings have been updated to include a `nugget` option.

### DIFF
--- a/MParT/MapFactory.h
+++ b/MParT/MapFactory.h
@@ -28,6 +28,7 @@ namespace mpart{
     options.quadRelTol = 1e-6;                          // Optional. Default = 1e-6
     options.quadMaxSub = 10;                            // Optional. Default = 30
     options.contDeriv = true;                           // Optional. Default = true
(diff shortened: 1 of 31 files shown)


dannys4 (approved):
Looks great! Would you mind putting an issue up on MParT.jl asking to include this in the MapOptions binding? I can take care of it from the Julia end.

mparno:
[MParT.jl issue #16](https://github.com/MeasureTransport/MParT.jl/issues/16) created.

PR #355, opened 2023-10-24
PyTorch Interface

mparno:
- Allows an MParT `ParameterizedFunctionBase` or `ConditionalMapBase` object to be used as a module in torch.  This allows torch autograd to propagate gradients through maps and allows neural networks and transport maps to be blended into more complex models.
- Importantly, we do not require pytorch during compilation.  The MParT <-> PyTorch interface will only be include with the python bindings if pytorch if available at the time of `import`
- Pytorch tensors are not copied when calling MParT.  We simply wrap unmanaged Kokkos::Views around the memory in the pytorch tensor.
- Currently only supports pytorch tensors that live on the CPU, but could be extended to work with GPU tensors as well.

- An example using this interface for heteroscedastic regression is in an[ accompanying PR into MParT-examples (PR 46)](https://github.com/MeasureTransport/MParT-examples/pull/46)

### DIFF
--- a/.github/environment.yml
+++ b/.github/environment.yml
@@ -12,5 +12,5 @@ dependencies:
   - pybind11 >=2.10
   - cereal >= 1.3
   - nlopt >= 2.7
-  - torch
+  - pytorch
 
--- a/.github/environment.yml
+++ b/.github/environment.yml
@@ -12,4 +12,5 @@ dependencies:
   - pybind11 >=2.10
   - cereal >= 1.3
   - nlopt >= 2.7
+  - torch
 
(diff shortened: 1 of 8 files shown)


dannys4:
Pinging @michael-c-brennan , just so he's aware of this. Notably, this is letting you integrate MParT *into* PyTorch, not the other way around.

Still remarkable!

dannys4:
@mparno do we want to add torch to the bindings test? It already takes forever :laughing:, adding another dependency for that test *probably* will do more good than harm.

dannys4:
I'd also be happy adding it here and commandeering the PR, but I *think* you just have to add `pytorch` to the list of conda-forge deps in `environment.yml`

mparno:
@dannys4 Good idea on the torch tests.  I just pushed up an updated `environment.yml`.

dannys4 (approved):
LGTM

PR #368, opened 2024-01-05
Add affine functions, edge terms to sigmoid basis

### LINKED ISSUE #366: Create affine fallback for SIgmoid
While working on #358 , I realized that Sigmoid was lacking a fallback to an affine function for order zero (which should be constant) and one (which should be linear). I thought about just fixing it inside the issue, but it would take even more revisions to tests and can easily be done outside of the PR. This should be pretty easy, the hardest part will be looking through my testing code.

dannys4:
In line with literature, I add affine functions to the `Sigmoid1d` class. Should be merged after #358 . Sorry to put this through so many changes, but that's why sigmoid is still only on `main`, I guess.

Fixes #366 . Now, if `s_j` is the `j`th order Sigmoid basis function, `s_0(x) = 1` and `s_1(x) = x`. This is backed up by the documentation added in #358 .


MParT/Sigmoid.h
```diff
@@ -1,245 +1,315 @@
...
+      ss << "can only evaluate up to order " << order_;
+      ProcAgnosticError<MemorySpace, std::invalid_argument>::error(
+          ss.str().c_str());
+    }
+
+    output[0] = 1.;
+    if (max_order == 0) return;
+    output[1] = input;
+    if (max_order == 1) return;
+		output[2] = -weights_(0)*EdgeType::Evaluate(-widths_(0)*(input-centers_(0)));
```
mparno, line 137: Is there funky indentation here, or is it just github rendering things funkily?

MParT/Sigmoid.h
```diff
@@ -1,245 +1,315 @@
...
-                param_idx++;
-            }
-        }
+    output[0] = 1.;
+    output_diff[0] = 0.;
+    if (max_order == 0) return;
+    output[1] = input;
+    output_diff[1] = 1.;
+    if (max_order == 1) return;
+		output[2] = -weights_(0)*EdgeType::Evaluate(-widths_(0)*(input-centers_(0)));
```
mparno, line 180: Same indentation question.

MParT/Sigmoid.h
```diff
@@ -1,245 +1,315 @@
...
-        }
+    output[0] = 1.;
+    output_diff[0] = 0.;
+    output_diff2[0] = 0.;
+    if (max_order == 0) return;
+    output[1] = input;
+    output_diff[1] = 1.;
+    output_diff2[1] = 0.;
+    if (max_order == 1) return;
+		output[2] = -weights_(0)*EdgeType::Evaluate(-widths_(0)*(input-centers_(0)));
```
mparno, line 234: Funky indents?

mparno (approved):
Looks good Danny!  There are a couple spots with funky indentation (at least on github), but otherwise I think it's good to go.  Thanks!

dannys4:
In my defense, VS Code keeps messing them up.

mparno (approved):

PR #370, opened 2024-01-22
Add move coeffs positional argument for julia binding

dannys4:
This was omitted in previous versions.


dannys4:
No extra functionality in base library, will merge without review as long as build tests pass.

PR #371, opened 2024-01-28
Fixed ParameterizedFunctionBase GPU bugs

MMRROOO:


dannys4:
Make sure to do something like the following
```
> make -j10 &> out.txt
> vim out.txt
```
then search `out.txt` in vim for `SetCoeffs` using `/SetCoeffs`. It looks at first glance that you're forgetting a few `template<>` for the `SetCoeffs` definition in the `.cpp` source file.

src/ParameterizedFunctionBase.cpp
```diff
@@ -190,14 +190,14 @@ void SetCoeffsInternal(unsigned int numCoeffs,
     if(coeffsDest.is_allocated()){
         if(coeffsSrc.size() != numCoeffs){
             std::stringstream msg;
-            msg << "Error in ParameterizedFunctionBase<MemorySpace>::SetCoeffs.  Expected coefficient vector with size " << numCoeffs << ", but new coefficients have size " << coeffs.size() << ".";
+            msg << "Error in ParameterizedFunctionBase<MemorySpace>::SetCoeffs.  Expected coefficient vector with size " << numCoeffs << ", but new coefficients have size " << numCoeffs << ".";
```
dannys4, line 193: This should stay as-is . `numCoeffs` is the number of coefficients in the PFB object, but `coeffs` are the new coefficients we want to include.

src/ParameterizedFunctionBase.cpp
```diff
@@ -208,48 +208,54 @@ void ParameterizedFunctionBase<MemorySpace>::SetCoeffs(Kokkos::View<const double
...
 
-    if(coeffs.size() != numCoeffs){
+    if(coeffsSrc.size() != numCoeffs){
         std::stringstream msg;
-        msg << "Error in ParameterizedFunctionBase<MemorySpace>::WrapCoeffs.  Expected coefficient vector with size " << numCoeffs << ", but new coefficients have size " << coeffs.size() << ".";
+        msg << "Error in ParameterizedFunctionBase<MemorySpace>::WrapCoeffs.  Expected coefficient vector with size " << numCoeffs << ", but new coefficients have size " << numCoeffs << ".";
         throw std::invalid_argument(msg.str());
     }
-    this->savedCoeffs = coeffs;
+    Kokkos::deep_copy(coeffsDest, coeffsSrc);
```
dannys4, line 221: This is going to be a little more complicated: WrapCoeffs should basically only make a shallow copy of the coefficients you give me. An analogy would be to imagine the following C syntax
```
void foo(double *x, double *y) {
    x = y; // WrapCoeffs
    *x = *y; // SetCoeffs
}
```

Here, you should only be setting the address of `coeffsDest` to `coeffsSrc`. In that same vein, `WrapCoeffs` should probably only work when `coeffsSrc` and `coeffsDest` are in the same memory space (this begs the question if an internal function like this is actually necessary, which I didn't consider before).

src/ParameterizedFunctionBase.cpp
```diff
@@ -208,48 +208,54 @@ void ParameterizedFunctionBase<MemorySpace>::SetCoeffs(Kokkos::View<const double
...
+template<typename MemorySpaceSrc, typename MemorySpaceDest>
+void WrapCoeffsInternal(unsigned int numCoeffs, 
+	Kokkos::View<double*,MemorySpaceDest>& coeffsDest,
+        Kokkos::View<double*,MemorySpaceSrc> coeffsSrc){
 
-    if(coeffs.size() != numCoeffs){
+    if(coeffsSrc.size() != numCoeffs){
         std::stringstream msg;
-        msg << "Error in ParameterizedFunctionBase<MemorySpace>::WrapCoeffs.  Expected coefficient vector with size " << numCoeffs << ", but new coefficients have size " << coeffs.size() << ".";
+        msg << "Error in ParameterizedFunctionBase<MemorySpace>::WrapCoeffs.  Expected coefficient vector with size " << numCoeffs << ", but new coefficients have size " << numCoeffs << ".";
```
dannys4, line 218: See comment above

dannys4 (changes requested):

dannys4 (approved):
lgtm

PR #372, opened 2024-01-29
Rectified mv expansion worker

dannys4:
Not glamorous at all, but this adds mechanisms with the end goal of creating a ConditionalMapBase linear on the coefficients of the following form: 

$$T(x,y) = \sum_{ \alpha_x\in J_x } b_{\alpha_x} \psi_{\alpha_x}(x) + \sum_{\alpha\in J} c_\alpha r(\psi_{\alpha_x}(x))g_{\alpha_y}(y)$$

 where
- $J$ is our multi-index set (with $J_x$ being shorthand for all multi-indices independent of $y$)
- { $\psi_{\alpha_x}$ } is our multivariate basis in $x$ (e.g., Hermite polynomials/functions)
- { $g_j$ } is a collection of univariate monotone functions (e.g., Sigmoids)
- $r:\mathbb{R}\to\mathbb{R}^+$ is a function that maps from the real line to positive numbers
- $b_{\alpha_x}\in\mathbb{R}$ and $c_{\alpha} \geq 0$

These requirements maintain monotonicity while being linear in the coefficients and keeping cross-terms (i.e. it is not necessarily separable in $x,y$). While we could imitate something like this in the previous form using a Multivariate Expansion for each of the nonmonotone and monotone parts, it would require sacrificing linearity in the monotone coefficients $c_\alpha$. This PR templates the `MultivariateExpansionWorker` using a `Rectifier` (representing $r$) and modifies the behavior accordingly. If the Rectifier is the identity operator, note that this formulation reverts to a generic multivariate expansion (and thus all the code should stay the same). Otherwise, we apply this formulation using a lot of `if constexpr` expressions.

Should be merged after #358


MParT/MultivariateExpansionWorker.h
```diff
@@ -454,6 +416,136 @@ class MultivariateExpansionWorker
 
 private:
 
+    KOKKOS_FUNCTION double GetTermVal(unsigned int termInd, const double* polyCache) const {
```
mparno, line 419: Does this function assume an implicit ordering or the multiindex?  What happens if `nzDims(i)==dim_-1` before we get to the last nonzero entry `end_idx`?  It might be more robust to keep two "termVals": one for the component that will be rectified and one for the others.

MParT/MultivariateExpansionWorker.h
```diff
@@ -454,6 +416,136 @@ class MultivariateExpansionWorker
...
+        // Compute the value of this term in the expansion
+        double termVal = 1.0;
+        unsigned int end_idx = multiSet_.nzStarts(termInd+1)-1;
+        for(unsigned int i=multiSet_.nzStarts(termInd); i<end_idx; ++i)
+            termVal *= polyCache[startPos_(multiSet_.nzDims(i)) + multiSet_.nzOrders(i)];
+        
+        double lastVal = polyCache[startPos_(multiSet_.nzDims(end_idx)) + multiSet_.nzOrders(end_idx)];
+        if constexpr(!std::is_same_v<Rectifier,Identity>){
```
mparno, line 427: Why do we need to check this here?  I wouldn't expect any performance hit if we just called `Rectifier::Evaluate`, especially if `Rectifier::Evaluate` is an inline function.

MParT/MultivariateExpansionWorker.h
```diff
@@ -454,6 +416,136 @@ class MultivariateExpansionWorker
...
+        // Compute the value of this term in the expansion
+        double termVal = 1.0;
+        int end_idx = multiSet_.nzStarts(termInd + 1) - 1;
+        if(multiSet_.nzDims(end_idx)!=dim_-1) return 0.; // Value is zero if constant in last dimension
+        for (unsigned int i = multiSet_.nzStarts(termInd); i < end_idx; ++i) {
+            termVal *= polyCache[startPos_(multiSet_.nzDims(i)) + multiSet_.nzOrders(i)];
+        }
+        double diagVal = polyCache[startPos_(posIndex) + multiSet_.nzOrders(end_idx)];
+
+        if constexpr (!std::is_same_v<Rectifier, Identity>) {
```
mparno, line 451: Same comment as above, not sure we need the `constexpr`

MParT/MultivariateExpansionWorker.h
```diff
@@ -454,6 +416,136 @@ class MultivariateExpansionWorker
...
+        int d = multiSet_.nzDims(end_idx); // Index of last nonconstant dimension
+        double lastVal;
+        if(wrt == d) {
+            lastVal = 1;
+            wrtVal = polyCache[startPos_(d) + multiSet_.nzOrders(end_idx)];
+            wrtDeriv = polyCache[startPos_(dim_+d) + multiSet_.nzOrders(end_idx)];
+        } else {
+            lastVal = polyCache[startPos_(d) + multiSet_.nzOrders(end_idx)];
+        }
+        if constexpr (!std::is_same_v<Rectifier, Identity>) {
```
mparno, line 482: Necessary?

MParT/MultivariateExpansionWorker.h
```diff
@@ -454,6 +416,136 @@ class MultivariateExpansionWorker
...
+                // wrt is the dimension we diff with respect to
+                wrtVal = polyCache[startPos_(wrt) + multiSet_.nzOrders(i)];
+                // We add dim_ to get the derivative index
+                wrtDeriv = polyCache[startPos_(dim_+wrt) + multiSet_.nzOrders(i)];
+            }else{
+                termVal *= polyCache[startPos_(multiSet_.nzDims(i)) + multiSet_.nzOrders(i)];
+            }
+        }
+        int d = multiSet_.nzDims(end_idx); // Index of last nonconstant dimension
+        double lastVal;
```
mparno, line 474: Same comment as above about the ordering of the nonzero dimensions.

MParT/MultivariateExpansionWorker.h
```diff
@@ -454,6 +416,136 @@ class MultivariateExpansionWorker
...
+        bool hasDeriv = false;
+        if(multiSet_.nzStarts(termInd)==multiSet_.nzStarts(termInd+1)) return 0.; // Value is zero if constant in last dimension
+        unsigned int end_idx = multiSet_.nzStarts(termInd+1)-1; // last index in loop
+        if(multiSet_.nzDims(end_idx)!=dim_-1) return 0.; // Value is zero if constant in last dimension
+        for(unsigned int i=multiSet_.nzStarts(termInd); i<end_idx; ++i){
+            termVal *= cache[startPos_(multiSet_.nzDims(i)) + multiSet_.nzOrders(i)];
+        }
+        double diagVal = cache[startPos_(posIndex) + multiSet_.nzOrders(end_idx)];
+        
+        if constexpr(!std::is_same_v<Rectifier,Identity>){
```
mparno, line 541: Needed?

MParT/MultivariateExpansionWorker.h
```diff
@@ -454,6 +416,136 @@ class MultivariateExpansionWorker
...
+                wrtVal = polyCache[startPos_(wrt) + multiSet_.nzOrders(i)];
+                wrtDeriv = polyCache[startPos_(dim_+wrt) + multiSet_.nzOrders(i)];
+            }else{
+                termVal *= polyCache[startPos_(multiSet_.nzDims(i)) + multiSet_.nzOrders(i)];
+            }
+        }
+        int posInd = (wrt==int(dim_)-1) ? (2*dim_) : (2*dim_-1);
+        double diagVal = polyCache[startPos_(posInd) + multiSet_.nzOrders(end_idx)];
+
+        if constexpr(!std::is_same_v<Rectifier,Identity>){
```
mparno, line 517: Needed?

MParT/MultivariateExpansionWorker.h
```diff
@@ -454,6 +416,136 @@ class MultivariateExpansionWorker
...
+    }
+
+    KOKKOS_FUNCTION double GetTermValMixedDeriv(unsigned int termInd, int wrt, const double* polyCache) const {
+        // Compute the value of this term in the expansion
+        double termVal = 1.0;
+        double wrtVal = 1.; // value of univariate basis fcn in dim wrt
+        double wrtDeriv = (wrt == int(dim_)-1 || wrt == -1); // derivative of univariate basis fcn in dim wrt
+        unsigned int end_idx = multiSet_.nzStarts(termInd+1)-1; // last index in loop
+        
+        if(multiSet_.nzDims(end_idx)!=dim_-1) return 0.; // Value is zero if constant in last dimension
```
mparno, line 504: Same concern about order dependence.  Maybe we should just make sure in the compressed multiindex that the nonzero dimensions are sorted.

mparno (commented):
This is going to be great!  I have two concerns that may or may not be valid but we should consider before merging:
- It seems like you require the `nzDims` array in the compressed multiindexset to be sorted.  Is that right?  If so, we should ensure that's case because I don't think in general we have any requirements for this to always be true even if it seems to work with our current multiindex factory methods.
- I'm not sure you need the `constexpr` statements to check if the rectifier type is the identity.   If it is the identity, wouldn't the compiler recognize that the static `Evaluate` method is just the identity?  I think it would, especially if we made that function inline.

tests/Test_MultivariateExpansionWorker.cpp
```diff
@@ -65,37 +83,37 @@ TEMPLATE_TEST_CASE( "Testing multivariate expansion worker", "[MultivariateExpan
     pt(1) = 0.1;
     pt(2) = 0.345;
 
-    // Fill in the cache the first d-1 components of the cache
+        // Fill in the cache the first d-1 components of the cache
     expansion.FillCache1(&cache[0], pt, DerivativeFlags::None);
-    double out[maxDegree+1];
+        double out[maxDegree+1];
```
mparno, line 88: Funky indent?

mparno:
Thanks for clarifying Dann.  I'm good with keeping the `constexpr` statements given your points about speed of the integrals.  

As for the multiindices, I think we should just make sure the dimensions are appropriately sorted in the `FixedMultiIndexSet` class.   I'll add this today.

mparno:
Sorting of `nzDims` added in #376

mparno (approved):

PR #373, opened 2024-01-30
ConditionalMapBase gpu bug fix

MMRROOO:


MParT/Utilities/ArrayConversions.h
```diff
@@ -320,6 +320,28 @@ namespace mpart{
 
     }
 
+    template<typename DeviceMemoryType, typename... OtherTraits>
```
dannys4, line 323: I don't think this function is needed anymore

dannys4 (changes requested):
Besides the extra function, this is looking good.

MParT/Utilities/ArrayConversions.h
MMRROOO: oops, I forgot to add that file to the commit

PR #376, opened 2024-01-31
Added sort of nzDims in manual construction of FixedMultiIndexSet.

mparno:
Ensures FixedMultiIndexSets have the order required in #372 .   To make sure everything can run on host or device, this usess an implementation of selection sort inside a functor evaluated by `Kokkos::parallel_for`.    Any speed savings possible using `std::sort` or something more sophisticated is probably negligible for the dimensions we typically work with.


dannys4 (approved):
With insertion sort, this looks good to me (runs well on my machine again)

PR #377, opened 2024-01-31
Change AddInPlace to lambda

dannys4:
Same functionality as before, but addresses issue seen in #375

Might be a problem with NVCC, Kokkos, or their integration? While syntactically this is equivalent to using `std::enable_if`, one of them compiles on the GPU and one doesn't, so I'm going with the one that works.


mparno (approved):
Looks good!  Thanks for figuring this out.

PR #380, opened 2024-01-31
Inner Marginal affine map

dannys4:
Representing a map $T_{MA}(x) = T(Dx + b)$, where $T:\mathbb{R}^{n}\to\mathbb{R}^{m}$ is a monotone map, $D\in\mathbb{R}^{n\times n}$ and $b\in\mathbb{R}^n$.

I know we talked a bunch about what to call this; I have unilaterally decided on `InnerMarginalAffineMap` because I wrote it :grin: . More precisely, this would be something like `IsotropicAffineOnceComposedMap` or `InnerIsotropicAffineMap` or something, but this is kinda precise and less wordy. The idea here is to use this for maps-from-samples so that, once a map is trained on standardized samples, you compose it with this class so you don't worry about standardization or whatever (you could compose it before training and it'd work fine, but you'd basically re-standardize every time you evaluate, which seems like a waste).


mparno (commented):
I like the idea of simplifying common use cases, but could you comment on why this is necessary instead of using the `ComposedMap` class with an `AffineMap`?

dannys4:
Yeah, so this isn't strictly "necessary" so-to-speak (as far as anything is necessary), but there's a few improvements.
1) this is an isotropic transformation, which means we don't need to use BLAS (i.e. no LU factorizations, matrix-matrix multiplication becomes an $O(Nd)$ operation instead of $O(Nd^2)$ operation for dimension $d$ and sample size $N$, inverses are done in quadratic instead of cubic time)
2) Checkpointing allocated extra memory that will not be strictly necessary. In particular, if you want to invert a composedmap, you have to do an extra allocation of the size $d \times N$. This isn't particularly concerning to me, to be honest, but it's a benefit.
3) This works with rectangular maps! ComposedMap doesn't.

PR #381, opened 2024-02-01
Kokkos update

mparno:
Changes to support Kokkos 4.  This version of kokkos no longer supports volatile arguments to lambda functions and I therefore replaced the `ReduceDim` class with direct uses of nested `parallel_for` and `parallel_reduce` calls.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -88,7 +88,7 @@ if(NOT Kokkos_FOUND)
         FetchContent_Declare(
             kokkos
             GIT_REPOSITORY https://github.com/kokkos/kokkos
-            GIT_TAG 3.7.00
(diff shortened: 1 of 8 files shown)


dannys4:
Odd. On my computer, when I try to use cmake FetchResources to install kokkos, I get that kokkos expects an lvalue for `teamMember` in `MapObjective.cpp:72`. Adding the extra `&` fixes it though. There's also an issue with the Julia bindings (the CI doesn't show that because of versioning issues).

dannys4 (approved):

dannys4:
Neat trick-- we can now build with c++20 (kokkos was what was holding us back), which allows for [some interesting flexibility](https://github.com/orgs/MeasureTransport/discussions/356).

PR #383, opened 2024-02-01
Added Diagonal Coefficient Functionality

dannys4:
A lot of algorithms explicitly depend on understanding which coefficients are associated with functions that are constant in the diagonal inputs and which coefficients are associated with functions that vary along diagonal inputs. Here I mean "diagonal" to say that, if coefficient $c_{\vec{\alpha}}$ (with associated multi-index $\vec{\alpha}$) is associated with output $j$, then $\alpha_j\neq 0$. This is useful for, e.g.
- When we restrict this set of coefficients to be nonnegative
- When performing Optimization routines

This PR adds a few nice tools for working with multi-indices in this sense and propagating some functionality through PFB. In particular,
- `Unfix` is like `Fix`, but the opposite (i.e. `mset.Fix(true).Unfix()` should  return a multi-index set equivalent to `mset`)
- `FixedMultiIndexSet::NonzeroDiagonalEntries`, which allows you to figure out which multi-indices have nonzero elements in their last entry
- `FixedMultiIndexSet::RemoveZeroDiagonalEntries`, which creates a new multi-index set from entries strictly nonzero in the last element.
- `NonzeroDiagTotalOrderLimiter`, which creates a TO mset but discards any elements with a zero at the end.
- `DiagonalCoeffIndices` for `ParameterizedFunctionBase`, with specializations for `MonotoneComponent` and `TriangularMap`. Default throws an error, but the specializations should give (linear) indices of coefficients in the map that correspond to basis functions that vary with respect to their last input.

### DIFF
--- a/MParT/MultiIndices/MultiIndex.h
+++ b/MParT/MultiIndices/MultiIndex.h
@@ -50,6 +50,19 @@ public:
         }
     }
 
+    /**
+     * @brief Construct a new Multi Index object from a set of nonzero indices and values
+     * 
+     * @param nzIndsIn indices of the nonzero values
+     * @param nzValsIn values that are nonzero
+     * @param numNz number of nonzero values (i.e., length of nzIndsIn and nzValsIn)
+     * @param lengthIn dimension of the index (numNz==lengthIn iff all values are nonzero)
+     */
+    MultiIndex(unsigned int* nzIndsIn,
+        unsigned int* nzValsIn,
+        unsigned int numNz,
+        unsigned int lengthIn);
+    
     /** Uses a dense vector description of the multiindex, defined through a pointer,
         and extracts the nonzero components.
         @param[in] fullVec A pointer the memory containing the dense multiindex.
--- a/MParT/MultiIndices/MultiIndex.h
+++ b/MParT/MultiIndices/MultiIndex.h
@@ -6,6 +6,8 @@
 #include <vector>
 
 #include <Eigen/Core>
+#include <Kokkos_Core.hpp>
+#include <iostream>
 
 namespace mpart {
 
@@ -55,14 +57,12 @@ public:
      * 
      * @param nzIndsIn indices of the nonzero values
      * @param nzValsIn values that are nonzero
-     * @param numNz number of nonzero values (i.e., length of nzIndsIn and nzValsIn)
      * @param lengthIn dimension of the index (numNz==lengthIn iff all values are nonzero)
      */
-    MultiIndex(unsigned int* nzIndsIn,
-        unsigned int* nzValsIn,
-        unsigned int numNz,
+    MultiIndex(Kokkos::View<unsigned int*, Kokkos::HostSpace> const& nzIndsIn,
+        Kokkos::View<unsigned int*, Kokkos::HostSpace> const& nzValsIn,
         unsigned int lengthIn);
-    
+
     /** Uses a dense vector description of the multiindex, defined through a pointer,
         and extracts the nonzero components.
         @param[in] fullVec A pointer the memory containing the dense multiindex.
--- a/MParT/ParameterizedFunctionBase.h
+++ b/MParT/ParameterizedFunctionBase.h
@@ -186,13 +186,6 @@ namespace mpart {
         /** Checks to see if the coefficients have been initialized yet, returns true if so, false if not */
         bool CheckCoefficients() const;
 
-        /**
-         * @brief Returns the indices of coeffs for functions non-constant in the last input for their respective output.
-         * 
-         * @return std::vector<unsigned int> 
-         */
-        virtual std::vector<unsigned int> DiagonalCoeffIndices() const;
-
         const unsigned int inputDim; /// The total dimension of the input N+M
         const unsigned int outputDim; /// The output dimension M
         const unsigned int numCoeffs; /// The number of coefficients used to parameterize this map.
--- a/MParT/ParameterizedFunctionBase.h
+++ b/MParT/ParameterizedFunctionBase.h
@@ -186,6 +186,13 @@ namespace mpart {
         /** Checks to see if the coefficients have been initialized yet, returns true if so, false if not */
         bool CheckCoefficients() const;
 
+        /**
+         * @brief Returns the indices of coeffs for functions non-constant in the last input for their respective output.
+         * 
+         * @return std::vector<unsigned int> 
+         */
+        virtual std::vector<unsigned int> DiagonalCoeffIndices() const;
+
         const unsigned int inputDim; /// The total dimension of the input N+M
         const unsigned int outputDim; /// The output dimension M
         const unsigned int numCoeffs; /// The number of coefficients used to parameterize this map.
--- a/src/MultiIndices/FixedMultiIndexSet.cpp
+++ b/src/MultiIndices/FixedMultiIndexSet.cpp
@@ -347,10 +347,9 @@ MultiIndexSet FixedMultiIndexSet<MemorySpace>::Unfix() const
     for(int term = 0; term < h_nzStarts.extent(0)-1; term++){
         unsigned int start = h_nzStarts(term);
         unsigned int end = h_nzStarts(term+1);
-        unsigned int numNz = end - start;
-        unsigned int* nzIndTerm = h_nzDims.data() + start;
-        unsigned int* nzValTerm = h_nzOrders.data() + start;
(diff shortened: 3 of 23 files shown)


MParT/ParameterizedFunctionBase.h
```diff
@@ -186,6 +186,13 @@ namespace mpart {
         /** Checks to see if the coefficients have been initialized yet, returns true if so, false if not */
         bool CheckCoefficients() const;
 
+        /**
+         * @brief Returns the indices of coeffs for functions non-constant in the last input for their respective output.
+         * 
+         * @return std::vector<unsigned int> 
+         */
+        virtual std::vector<unsigned int> DiagonalCoeffIndices() const;
```
mparno, line 194: I'm not sure it makes sense to expose this in `ParameterizedFunctionBase` because the function could be rectangular, is generally not triangular, and the definition of what is the "diagonal" might not be as well-defined as it is for `MonotoneComponent` or `TriangularMap`.   Is there anywhere we need this in PFB or can we always cast to a child with a well-defined diagonal?

src/MultiIndices/FixedMultiIndexSet.cpp
```diff
@@ -244,6 +245,25 @@ std::vector<unsigned int> FixedMultiIndexSet<Kokkos::HostSpace>::IndexToMulti(un
 }
 
 
+template<typename MemorySpace>
+std::vector<unsigned int> FixedMultiIndexSet<MemorySpace>::NonzeroDiagonalEntries() const
+{
+    assert(false);
```
mparno, line 251: Can we also implement this for devices?

src/MultiIndices/FixedMultiIndexSet.cpp
```diff
@@ -244,6 +245,25 @@ std::vector<unsigned int> FixedMultiIndexSet<Kokkos::HostSpace>::IndexToMulti(un
...
+    return std::vector<unsigned int>();
+}
+
+template<>
+std::vector<unsigned int> FixedMultiIndexSet<Kokkos::HostSpace>::NonzeroDiagonalEntries() const
+{
+    std::vector<unsigned int> output;
+    if(!isCompressed) throw std::runtime_error("NonzeroDiagonalEntries only works for compressed multiindex sets");
```
mparno, line 259: I think this function should work whether or not the `FixedMultiIndex` is compressed.  If compressed, can we just revert to a nested loop?  It would be slower, but would work.

MParT/MultiIndices/MultiIndex.h
```diff
@@ -50,6 +50,19 @@ class MultiIndex {
...
 
+    /**
+     * @brief Construct a new Multi Index object from a set of nonzero indices and values
+     * 
+     * @param nzIndsIn indices of the nonzero values
+     * @param nzValsIn values that are nonzero
+     * @param numNz number of nonzero values (i.e., length of nzIndsIn and nzValsIn)
+     * @param lengthIn dimension of the index (numNz==lengthIn iff all values are nonzero)
+     */
+    MultiIndex(unsigned int* nzIndsIn,
```
mparno, line 61: I think this would be safer if it didn't use raw pointers.  Instead, could we use Kokkos views and use a subview when we call this constructor in `FixedMultiIndexSet::Unfix`?

src/MultiIndices/FixedMultiIndexSet.cpp
```diff
@@ -299,6 +319,27 @@ int FixedMultiIndexSet<MemorySpace>::MultiToIndex(std::vector<unsigned int> cons
...
+    Kokkos::View<unsigned int*, Kokkos::HostSpace> h_nzDims = Kokkos::create_mirror_view_and_copy(Kokkos::HostSpace(), nzDims);
+    Kokkos::View<unsigned int*, Kokkos::HostSpace> h_nzOrders = Kokkos::create_mirror_view_and_copy(Kokkos::HostSpace(), nzOrders);
+    MultiIndexSet output (this->dim);
+    for(int term = 0; term < h_nzStarts.extent(0)-1; term++){
+        unsigned int start = h_nzStarts(term);
+        unsigned int end = h_nzStarts(term+1);
+        unsigned int numNz = end - start;
+        unsigned int* nzIndTerm = h_nzDims.data() + start;
+        unsigned int* nzValTerm = h_nzOrders.data() + start;
+        MultiIndex midx_term {nzIndTerm, nzValTerm, numNz, this->dim};
```
mparno, line 337: See comment above about using subviews instead of pointers, which I think would be safer without sacrificing much speed.

src/ParameterizedFunctionBase.cpp
```diff
@@ -21,6 +21,13 @@ void ParameterizedFunctionBase<MemorySpace>::CheckDeviceMismatch(std::string fun
     throw std::runtime_error(msg.str());
 }
 
+template<typename MemorySpace>
+std::vector<unsigned int> ParameterizedFunctionBase<MemorySpace>::DiagonalCoeffIndices() const
+{
+    throw std::logic_error("DiagonalCoeffIndices not implemented for the given Parameterized Function class.");
```
mparno, line 27: Maybe another reason for not exposing this in PFB.

mparno (changes requested):
Thanks for these additions!  I think it's really close, but there are a few corner cases that I think we should finish up implementing and I have a question about using raw pointers.

MParT/ParameterizedFunctionBase.h
dannys4: In C++ land, you're likely correct. However, in Julia/Python/Matlab land (or at least julia-land), the answer is a resounding "unsure". There's actually a bigger corner case of `ComposedMap`, which is just a disaster in- and of- itself (Suppose you compose a triangular map with a sparse affine map; then the diagonal indices would come from some graph algorithm). I agree with you that this is not a "good" way of exposing this functionality. On the other hand, a) since this doesn't make sense for nontriangular maps, I don't see the actual damage; b) I suspect there are very few working with "rectangular" maps that aren't composed; and c) working with bindings makes any "downcasting" from conditional map base questionably difficult and I'm not enthusiastic about an exercise in now needing to cast maps between different levels in the class hierarchy for different functionality as well as getting that working in three binding languages.

src/MultiIndices/FixedMultiIndexSet.cpp
dannys4: Yeah, that's fine. I just didn't know what an uncompressed fmset looks like; there's no documentation on it and I wasn't sure where it was used. I figured it was better to write a function that succeeds on most cases then explicitly fails for sure on something that I don't understand than write a function that might or might not fail due to my lack of understanding of the corner case.

dannys4:
Okay, I pushed up some changes resolving the safety and implementation of multi-index methods. Let me know what you think about the `PFB::DiagonalCoeffIndices`.
Two other resolutions I can imagine:
- only allow it for PFB of output-size 1 where we implement it as something like `LastInputCoefficients` and save the actual diagonalcoeffindices for TriangularMap
- In the bindings, bind it to PFB but then bind it via lambda with pseudocode like
```c++
std::shared_ptr<TriangularMap> pfb_cast = cast<TriangularMap>(pfb_pointer);
if(pfb_cast == nullptr) {throw error}
else {return pfb_cast->DiagonalCoeffIndices;}
```
(obviously this basically the same solution, but keeps the C++ base cleaner while pawning the "unsafety" off to the 3 binding implementations)

MParT/ParameterizedFunctionBase.h
mparno: I don't really like the idea of adding things to the base class just out of convenience for the bindings, but if it would be a royal pain otherwise than I could be convinced.  At least for python, I don't really see any issues.  pybind11 is very good at handling these things and figuring our what child type the objective is.

src/MultiIndices/FixedMultiIndexSet.cpp
mparno: If I remember correctly, the "compressed" flag just indicates that any 0.0 values in the array of nonzeros have been removed.

dannys4: Should've marked this as resolved; it's been added!

dannys4:
Sorry to sneak in a small fix after review, but I realized I actually didn't bind `DiagonalCoeffIndices` in this PR, so the effects of moving the hierarchy around is minimal, at least here.

mparno (approved):
Thanks for working with Danny.  I'm ready to merge when you are.

PR #384, opened 2024-02-02
CreateSigmoidExpansion

dannys4:
This creates a factory method, `CreateSigmoidExpansion`, which does exactly what it sounds like. To be clear, this is not a useful function as-is (stated in the included documentation), because it only uses a rectified BasisEvaluator; i.e. this can only evaluate a multivariate expansion of the form
$$T(x,y) = \sum_{\vec{\alpha}} c_{\vec{\alpha}} r(\Psi_{\vec{\alpha}(x)}(x))s_{\alpha(y)}(y)$$

where $r:\mathbb{R}\to\mathbb{R}^+$, $\vec{\alpha}$ is our multi-index with off-diagonal and diagonal elements corresponding to $x$ and $y$ respectively, and all coefficients with $\vec{\alpha}(y)$ should be positive. What this means is that even the elements of the expansion that do *not* correspond to $y$ are always one sign. This *might* be useful; I genuinely have no idea. However, this is a stepping stone PR to create an easy-to-use `RectifiedMultivariateExpansion` by creating much of the software machinery for it.

Note that this has breaking changes for `MapOptions`, which means that you won't be able to deserialize options from a previous version within this version; I'm not super happy about this, but this is a compromise that doesn't seem easy to avoid. We've already created a type `MapOptions`, which seems to subsume all possible options a map should have from a UX perspective. If I create a new map (not based on the MonotoneComponent) that has its own options, surely `MapOptions` should work to construct that. Otherwise, what's the point? So I'm not really happy about creating a new option type just for a different kind of map to tell our users that "We have maps, but also we have these other things that are also maps but they take different options even though they're the same thing conceptually". On the other hand, renaming it is a bad decision because then we'd have to change basically everything. This is minimally invasive, though unfortunately it is possibly annoying. However, it's easy to print out mapoptions in human-readable format, so porting any serialized version should be easy.

### DIFF
--- a/MParT/MapFactory.h
+++ b/MParT/MapFactory.h
@@ -105,20 +105,41 @@ namespace mpart{
 
 
         /**
-        @brief Constructs a (generally) non-monotone multivariate expansion.
-        @param outputDim The output dimension of the expansion.  Each output will be defined by the same multiindex set but will have different coefficients.
-        @param mset The multiindex set specifying which terms should be used in the multivariate expansion.
-        @param options Options specifying the 1d basis functions used in the parameterization.
+        @brief Constructs a "Single entry" map, which is an identity for all but one output dimension.  The active output dimension is defined by a given component.
+        @param dim The dimension of the mapping
+        @param activeInd Which output dimension is "active", i.e., not identity
+        @param comp The component that will be used for the active output dimension.
         */
         template<typename MemorySpace>
         std::shared_ptr<ConditionalMapBase<MemorySpace>> CreateSingleEntryMap(unsigned int dim,
-                                                                                     unsigned int activeInd,
-                                                                                     std::shared_ptr<ConditionalMapBase<MemorySpace>> const &comp);
+                                                                              unsigned int activeInd,
+                                                                              std::shared_ptr<ConditionalMapBase<MemorySpace>> const &comp);
 
+        /**
+         * @brief Create a MultivariateExpansion using a sigmoid object; INTERNAL USE ONLY
+         * 
+         * @details Using the parameters described in \c`opts`, this function creates
+         * a "multivariate expansion" object that uses a collection of sigmoid functions
+         * to map the input space to the output space. The pertinent options are
+         * - \c`opts.basisType` : The type of basis function to use in the expansion for the first \f$d-1\f$ inputs
+         * - \c`opts.posFuncType` : The type of positive function to use in the rectified basis @ref MultivariateExpansionWorker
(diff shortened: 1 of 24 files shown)


dannys4:
For discussion's sake, I'll note that I tried coalescing the two possibilities, but this would be somewhat difficult apriori without adding some kind of versioning tooling to all serialization (which would then just break all serialization in total). Further, due to the structure of the Mex+ bindings for MapOptions, there is no reasonable solution that doesn't break matlab (matlab needs to know the number of outputs a priori, and while I can imagine a solution, it would be far more onerous and complicated. In particular, it would not hold up to further adaptations of the classes).

MParT/MapFactory.h
```diff
@@ -105,15 +105,41 @@ namespace mpart{
...
+         * @details Using the parameters described in \c`opts`, this function creates
+         * a "multivariate expansion" object that uses a collection of sigmoid functions
+         * to map the input space to the output space. The pertinent options are
+         * - \c`opts.basisType` : The type of basis function to use in the expansion for the first \f$d-1\f$ inputs
+         * - \c`opts.posFuncType` : The type of positive function to use in the rectified basis @ref MultivariateExpansionWorker
+         * - \c`opts.edgeWidth` : The width of the "edge terms" on the last input @ref Sigmoid1d
+         * - \c`opts.sigmoidType` : The type of sigmoid function to use in the expansion @ref Sigmoid1d
+         * 
+         * By default, this constructs a total-order multi-index set. Note that, since the basis
+         * evaluator is rectified, this *will always* be a positive-valued function in the first \f$d-1\f$ inputs.
```
mparno, line 130: Does #383 allow us to lift this restriction?

mparno (commented):
Hey Danny!  Thanks for pulling this together and especially for cleaning up the matlab bindings.  The code looks great, but based on your comments and the restrictions that even off-diagonal terms are positive, I wonder if this is ready to be merged into main.  If #383 allows us to lift this off-diagonal-positive restriction, would you be open to incorporating that and then merging into main?  I hesitate to introduce too many "works in progress" into main because it can lead to having a lot of partially-implemented features that can turn into "noise" in the released package and detract from the more polished tools.

mparno (approved):

MParT/MapFactory.h
dannys4: Unfortunately no, this is actually changed drastically in #385/#386 (yes, a `dev` branch might be in order here).

PR #385, opened 2024-02-05
Rectified MVE class

dannys4:
Exactly what it sounds like; a class, as described in previous PRs, to define a function $T:\mathbb{R}^d\times\mathbb{R}\to\mathbb{R}$ which performs
$$T(\mathbf{x},y) = \sum_{\vec{\alpha}^1_x} c^1_{\vec{\alpha}^1_x} \Psi_{\vec{\alpha}^1_x}(\mathbf{x})+ \sum_{\vec{\alpha}^2}c^2_{\vec{\alpha}^2}r\left[\Psi_{\vec{\alpha}^2_x}(\mathbf{x})\right]g_{\alpha^2_y}(y),$$

where $\vec{\alpha}^1_x$ and $\vec{\alpha}^2$ do not necessarily come from the same multi-index set (indeed, we want all $\vec{\alpha}^2$ to come from a set that is exclusively non-constant in $y$ to avoid redundant terms); further, all $\{g_j\}$ should be monotone, and $r:\mathbb{R}\to\mathbb{R}^+$ should probably be bijective (not sure how necessary that is though), and the second set of coefficients, $\{c^2_{\vec{\alpha}_2}\}$, must be nonnegative.

This went through several iterations and I think this is the least bad choice here. The provided MVE would've worked to some degree, but defining the LogDet and Inverse functions would've been invasive, and we'd lose some efficiency from spawning more parallelism than necessary (as well as the fact that MVE is engineered to give many outputs simultaneously, whereas this is a special case).

Bindings and a MapFactory method will be created in a future PR.


MParT/DerivativeFlags.h
```diff
@@ -9,7 +9,7 @@ namespace DerivativeFlags{
         Parameters, //<- Deriv wrt coeffs
         Diagonal,   //<- first derivative wrt diagonal
         Diagonal2,  //<- second derivative wrt diagonal
-        Mixed,      //<- gradient wrt coeffs of first derivative wrt x_d
+        MixedCoeff, //<- gradient wrt coeffs of first derivative wrt x_d
```
mparno, line 12: Thanks for making this consistent!

MParT/MultivariateExpansion.h
```diff
@@ -93,9 +93,9 @@ namespace mpart{
             Kokkos::fence();
         }
 
-        virtual void GradientImpl(StridedMatrix<const double, MemorySpace> const& pts,  
-                                  StridedMatrix<const double, MemorySpace> const& sens,
-                                  StridedMatrix<double, MemorySpace>              output) override
+        void GradientImpl(StridedMatrix<const double, MemorySpace> const& pts,
```
mparno, line 96: While not strictly necessary, I would advocate for keeping the virtual keyword across all of the `*Impl` classes.  It avoids some compiler warnings and, even though we have the `override` keyword, makes it more obvious that these functions stem from the base class.

MParT/Utilities/RootFinding.h
```diff
@@ -105,11 +105,23 @@ KOKKOS_INLINE_FUNCTION double Find_x_ITP(double xlb, double xub, double yd, doub
...
+ * @tparam MemorySpace
+ * @tparam FunctorType
+ * @param yd Given output value
+ * @param f Functor
+ * @param x0 Initial guess
+ * @param xtol Tolerance for the root
+ * @param ftol Tolerance for the function value
+ * @param info Output flag
+ * @return root of the function \f$f(\cdot)-y_d\f$
+ */
```
mparno, line 124: Thanks for flushing out the documentation.

mparno (changes requested):
Thanks Danny!  I have one comment about the use of virtual keywords, but everything of real substance looks great!  It'll be great to have this in MParT.

mparno (approved):
Thanks Danny!

PR #386, opened 2024-02-07
Bindings for multi-index methods, factory methods and bindings for sigmoids, new 1d map

dannys4:
Builds on #384 #385 #383 for ultimate user experience.
- ✨**All new class**✨:`UnivariateExpansion` is exactly what it promises. A quick and dirty 1d expansion which just trusts you to use a monotone 1d basis (see where concepts/constraints could come in yet?)
- ✨**Fresh bindings**✨: ~~Bind `DiagonalCoeffIndices` for users to see which coefficients correspond to multivariate basis functions varying in $x_j$ for the $j$th input. Doesn't currently distinguish between inputs.~~ This will be pushed off to a new PR due to changes in #383 .
- ✨**New factory methods**✨: `CreateSigmoidComponent` and `CreateSigmoidTriangular` mimic their non-sigmoid parts, except they take in (fixed) parameters for where the centers of the sigmoids should be. They also automatically choose the bandwidth of the sigmoids according to a heuristic (TODO: allow the user to choose this from the bindings). As far as I understand, *where* you place the centers generally makes a larger difference than how wide the sigmoids are, within reason.

I tested all the bindings locally in all the languages, and everything seemed to work. I'm hoping that this closes out the slew of PRs that leads to MParT 3.0, though some documentation may be in order.


dannys4:
A very good criticism to consider that I just thought of is that there's no clear way here to get from Multi-index set to sigmoid component, which makes adaptive methods tricky. Due to the structure of the rectified MVE, you actually need two msets (satisfying some conditions), and so I'm not sure how accessible I should make this. Maybe I'll add a secondary backup method that takes two fixed msets?

bindings/matlab/tests/test-results-matlab.xml
mparno, line 1: Should this be in here?

mparno (changes requested):
Nice work Danny!  Another great addition.  There is a matlab test file that looks like it might have snuck in, but everything else looks good to me.

bindings/matlab/tests/test-results-matlab.xml
dannys4: oops

mparno (approved):

PR #389, opened 2024-02-14
GPU fully compiling

MMRROOO:


MParT/ParameterizedFunctionBase.h
```diff
@@ -51,6 +51,8 @@ namespace mpart {
             @param coeffs A view containing the coefficients to copy.
         */
        virtual void SetCoeffs(Kokkos::View<const double*, MemorySpace> coeffs);
+       virtual void SetCoeffs(Kokkos::View<double*, MemorySpace> coeffs);
```
dannys4, line 54: This is a good idea, but I think you should just define it as such:
```cpp
void SetCoeffs(Kokkos::View<double*, MemorySpace> coeffs) {
    Kokkos::View<const double*, MemorySpace> const_coeffs;
    SetCoeffs(const_coeffs);
}
```
The differences here:
- It is not a virtual function
- It allows us to ensure only one `SetCoeffs` needs to be implemented in child classes

MParT/Utilities/ArrayConversions.h
```diff
@@ -223,23 +223,10 @@ namespace mpart{
     /**
     @brief Copies a Kokkos array from device memory to host memory
     @details
-    @param[in] inview A kokkos array in device memory.
-    @return A kokkos array in host memory.  Note that the layout (row-major or col-major) might be different than the default on the Host.  The layout will match the device's default layout.
+    @tparam ViewType A Kokkos View type in device space.
+    @param[in] inview A kokkos view ind device space.
+    @return A kokkos array in host memory, that is a copy of the inview.
```
dannys4, line 228: You can get rid of this documentation

src/ParameterizedFunctionBase.cpp
```diff
@@ -208,6 +208,13 @@ void ParameterizedFunctionBase<MemorySpace>::SetCoeffs(Kokkos::View<const double
     SetCoeffsInternal(this->numCoeffs, this->savedCoeffs, coeffs);
 }
 
+template<typename MemorySpace>
+void ParameterizedFunctionBase<MemorySpace>::SetCoeffs(Kokkos::View<double*, MemorySpace> coeffs){
+    Kokkos::View<const double*, MemorySpace> tmpCoeffs = coeffs;
```
dannys4, line 213: See other comment

dannys4 (changes requested):
Largely great, keep up the good work. Should only be a few minor changes hopefully.

dannys4:
Last comment: it looks like, in the github actions, the new SetCoeffs breaks Julia due to function call ambiguity; this should be fixed by just separating out the lines like
```c++
Kokkos::View<const double*, MemorySpace> const_coeffs = JlToKokkos(coeffs);
pfb.SetCoeffs(const_coeffs);
```

PR #391, opened 2024-02-23
Re-adding GPU Operation

dannys4:
GPU works, just as before. Passes all tests and, locally, works with Python. Waiting on 384-386 to merge into main before this is ready for review (lots of merge conflicts incoming).


mparno (approved):
Awesome work @MMRROOO and @dannys4 !  I successfully compiled with with CUDA 12.0 on a machine with an NVIDIA RTX A5500 GPU.  It compiled and all the tests ran successfully!   

During compilation I saw a bunch of warnings like the following:
```bash
MParT/BasisEvaluator.h(321): warning #20011-D: calling a __host__ function("std::vector<    ::std::shared_ptr<    ::TestEvaluators> , ::std::allocator<    ::std::shared_ptr<    ::TestEvaluators> > > ::operator [](unsigned long) const") from a __host__ __device__ function("mpart::BasisEvaluator<( ::mpart::BasisHomogeneity)2,     ::std::vector<    ::std::shared_ptr<    ::TestEvaluators> , ::std::allocator<    ::std::shared_ptr<    ::TestEvaluators> > > ,  ::mpart::Identity> ::EvaluateSecondDerivatives const") is not allowed
```

I'm fine merging this PR as is, but noted that these warnings are something we should address in Issue #397

PR #401, opened 2024-03-27
Fix Issue with mpart.py when mset has one entry #286

### LINKED ISSUE #286: Issue with mpart.py when mset has one entry
In Python
```python
import mpart as mt
import numpy as np
multis = np.array([[1,]])
mset = mt.MultiIndexSet(multis)
mset.Size() # is zero
```
Interestingly, in Julia
```julia
using MParT
multis = reshape([1], 1, 1)
mset = MultiIndexSet(multis)
Size(mset) # is one
```
Haven't tried it in Matlab. But I believe, if I make an mset with one index, it should have one coefficient. Right? Why are these different?

dannys4:
Fix #286


bindings/python/src/MultiIndex.cpp
```diff
@@ -25,6 +25,44 @@
...
+    int stride_size = is_int32 ? sizeof(int32_t) : sizeof(int64_t);
+    Eigen::Stride<Eigen::Dynamic, Eigen::Dynamic> strides(
+        info.strides[rowMajor ? 0 : 1] / (py::ssize_t)stride_size,
+        info.strides[rowMajor ? 1 : 0] / (py::ssize_t)stride_size
+        );
+
+    if(is_int64) { // Is int64
+        Matrix_Map_T<int64_t> map (static_cast<int64_t*>(info.ptr), info.shape[0], info.shape[1], strides);
+        Eigen::Matrix<int32_t,Eigen::Dynamic,Eigen::Dynamic> mat_32t {info.shape[0], info.shape[1]};
+        for(int i = 0; i < info.shape[0]; i++) {
```
mparno, line 54: Instead of manually copying each entry, I think you could just do `mat_32t=map.cast<int32_t>()`, right?

mparno (changes requested):
One small recommendation, otherwise I think it's good to go.  Thanks @dannys4 !

dannys4:
Closing this with the intention of creating a new one to merge into `release`.

PR #404, opened 2024-04-01
Refactor Terminology for Consistency

dannys4:
Large shift of terminology here. Move everything outside of MVE to consistently use "parameter" terminology as well as moving `Gradient` to become `InputGrad`.

This hinges on `mpart::DeprecationWarning`, which will store what deprecated functions you call and the first few times (default 2) it will display a warning. However, this shouldn't spam the console for applications that call the same deprecated function(s) thousands of times due to the limit.

The deprecated functions automatically call their renamed siblings, so all behavior should be identical to previously.


dannys4 (commented):
These are going to be the changes in the end

- `LogDeterminantCoeffGrad`->`LogDeterminantParamGrad`
- `CoeffGrad`->`ParamGrad`
- `numCoeffs`->`numParams`
- `Coeffs`->`Params`
- `CoeffMap`->`ParamMap`
- `SetCoeffs`->`SetParams`
- `WrapCoeffs`->`WrapParams`
- `moveCoeff`->`moveParam`
- `Gradient`->`InputGrad`
- `MapObjective::CoeffGrad`->`ParamGrad`

mparno:
This is a herculean effort, thanks Danny!   I still need to double check that we've caught everything, but I really like the use of your deprecated function for warnings, especially because that will trickle up to the bindings.  I was wondering though, if you had any thoughts on also using the `[[deprecated]]` feature of c++14?  It seems like that this is a natural use case for it.

dannys4:
Closing this as unplanned in favor of better documentation of the current (possibly regrettable) choices made.

Issue #398, opened 2024-03-26
Preliminary training with GPU-acceleration


mparno:
Need to hook up host-only optimizer with possible device-accelerated map evaluation.

dannys4:
Added in #408

PR #408, opened 2024-04-04
Add Map Training Features

### LINKED ISSUE #334: Refine `TrainMap`
For rectified polynomials we can take advantage of the structure of the problem in a couple of ways:
- [x] We should take advantage of when we transport to a Normal distribution in our optimization by separating out the optimization objective by component so we solve `d` one dimensional problems instead of one `d` dimensional problem
- [x] In the KL case, we know the optimization objective is convex. Why not use, e.g., Nesterov accelerated gradient descent?

dannys4:
This adds a few new features,  resolves #334 #398
- Internal class `FastGaussianReverseKLObjective`. Exactly what it sounds like, it's a shortcut class that estimates the Reverse KL for when you are using the unshifted/scaled reverse KL optimization objective (i.e. map from samples)
    - This only works when the map has scalar-valued output, which ensures maximum speed. If you want flexibility, you can use the previouse `KLObjective`
    - The traditional KLObjective, when used in `TrainMap` on a `TriangularMap` with a Gaussian reference, will automatically be constructed into a sequence of these objects (using subviews of the training dataset of course). While there are certainly small optimizations to add here on memory reuse, this is a pretty flexible way of ensuring that any previous code using `TrainMap` will accelerate without any changes from the user.
    - Any bindings are outside the scope of the PR.
- Compatibility and testing of GPU-based map training
    - As discussed in #398, this just moves the data to and from device and does all optimization on the host (still using NLOpt). 
    - Adds tests for GPU-based map training as well, ensuring that anything we tested for the CPU in terms of training also works for the GPU.


dannys4:
Since I now subvert all default KL objectives from samples to this "fast" version, I just added a new test that ensures we use a proper pullback density as the objective (i.e., I construct the covariance of the Gaussian explicitly as identity, which forces it to behave as previously). For some reason, this doesn't train correctly *only* on CPU. For that reason, I think it's an issue with Eigen and I'm not really sure how to fix it. I would appreciate any help here.

mparno (approved):
Looks good to me.  Thanks Danny!   This is going to be a huge improvement in the training speed.

PR #410, opened 2024-04-24
Accelerate rectified multivariate expansion

### LINKED ISSUE #400: Accelerate `RectifiedMultivariateExpansion`
Currently, `RectifiedMultivariateExpansion` duplicates a lot of work because it represents a function of the form
$$T(\mathbf{x},y) = f(\mathbf{x}) + g(\mathbf{x},y)$$
by using $f,g$ as two separate objects. This is due to a constraint in the rectification for a basis evaluator. However, most of the time, $f$ and $g$ are made of multivariate polynomials and so the current implementation duplicates a lot of work, resulting in almost doubled runtime for multivariate bases with many cross-terms.

To fix this, we need to
- Add the ability to do rectification only on the terms depending on $y$ in the `MultivariateExpansionWorker`
- Change the interface for `RectifiedMultivariateExpansion` to only take in one `FixedMultiIndexSet`
- Change the evaluation for `RectifiedMultivariateExpansion` to only use one `MultivariateExpansionWorker`
- Change the all of the bindings and tests accordingly

dannys4:
Resolves #400 and also streamlines the construction/use of RMVE.


MParT/MultivariateExpansionWorker.h
```diff
@@ -483,13 +483,18 @@ class MultivariateExpansionWorker
             lastVal = polyCache[startPos_(d) + multiSet_.nzOrders(end_idx)];
         }
         if constexpr (!std::is_same_v<Rectifier, Identity>) {
+            bool isRectified = d == dim_ - 1;
             if(wrt == dim_ - 1) { // Diagonal deriv
                 termVal = Rectifier::Evaluate(termVal)*wrtDeriv; // if wrt != d, wrtDeriv = 0
             } else if (wrt == -1) { // No deriv
-                termVal = (d == dim_ - 1) ? Rectifier::Evaluate(termVal)*lastVal : Rectifier::Evaluate(termVal*lastVal);
+                termVal = (isRectified) ? Rectifier::Evaluate(termVal)*lastVal : termVal*lastVal;
```
mparno, line 490: I'm a little confused why we're using the rectifier here.  

### What I thought we were doing:
I was under the impression that the rectified expansion took the form 

$$
S(x; \mathbf{y}) = h(\mathbf{y}) + \sum_{i=0}^N c_i r( g_{i}(\mathbf{y})) s_i(x),
$$

where $h(\mathbf{y})$ and $g_i(\mathbf{y})$ are typical multivariate polynomial expansions with the form

$$
g_i(\mathbf{y}) = \sum_{\alpha\in\mathcal{A}} b_i \left(\prod_{d=1}^{D_y} \phi_d(y_d)\right)
$$

In this setting, the rectification is applied to the output of a multivariate polynomial expansion. 

### What this looks like:
I'm likely missing how this all comes together, but this function seems to imply that we might instead be defining expansions with the form 

$$
S(x; \mathbf{y}) = h(\mathbf{y}) + \sum_{\alpha \in \mathcal{A}} c_\alpha \left(s_{\alpha_{D_y+1}}(x) \prod_{d=1}^{D_y} r( \phi_{\alpha_{d}}(y_d) ) \right),
$$

In this setting the rectification is applied to the first $D_y$ components of the multivariate basis function, which spans a very different function space and will lose any orthogonality properties of the polynomial basis.

### Questions
1. Is my interpretation correct here?
2. If so, why is this the form?  Would it be better to leave the ExpansionWorker alone and use the first form described above?

MParT/RectifiedMultivariateExpansion.h
```diff
@@ -72,22 +64,15 @@ namespace mpart{
...
-                    // Evaluate the expansion
-                    output_slice(ptInd) = worker_off.Evaluate(cache.data(), coeff_off);
 
                     // Fill in entries in the cache that are dependent on x_d.  By passing DerivativeFlags::None, we are telling the expansion that no derivatives with wrt x_1,...x_{d-1} will be needed.
-                    worker_diag.FillCache1(cache.data(), pt, DerivativeFlags::None);
-                    worker_diag.FillCache2(cache.data(), pt, pt(pt.size()-1), DerivativeFlags::None);
-                    output_slice(ptInd) += worker_diag.Evaluate(cache.data(), coeff_diag);
+                    worker.FillCache1(cache.data(), pt, DerivativeFlags::None);
+                    worker.FillCache2(cache.data(), pt, pt(pt.size()-1), DerivativeFlags::None);
+                    output_slice(ptInd) = worker.Evaluate(cache.data(), coeff);
```
mparno, line 75: See my comment above about the form of this expansion.  I was expecting something very different.

mparno (commented):
Thanks Danny.! This is a big effort, especially with all the bindings.  Unfortunately the implementation does not conceptually match what I thought we were doing.  Could you clarify the mathematical form of the Rectified expansion?  I've tried to describe the difference between what I thought we were doing and what the implementation looks like in my comment below.  Thanks!

MParT/MultivariateExpansionWorker.h
dannys4: I should emphasize that this doesn't change the parameterization-- it seems like you may have misunderstood what it was doing before (see #385 for a writeup of how it worked before, which is close to what you write for "What this looks like"). The difference I will emphasize is that it's $r(\Phi_\alpha)=r(\prod \phi_{\alpha_d})$, *not* $\prod r(\phi_{\alpha_d})$

The point here is that, in two dimensions, we had to create two different MVE workers: one for multi-indices of the form $(\alpha_1,0)$ and one for multi-indices of the form $(\alpha_1,\alpha_2)$ for $\alpha_2 > 0$. The reason is the implementation of how the rectification worked before in the MVE worker class. This just changes it so that we don't need to recompute a bunch of polynomials and waste both time and space repeating work with MVE worker 1 vs MVE worker 2.

dannys4: If it makes you feel any better, for $T(x,y)$, fix a $y^\star$ and consider $\partial_y T(x,y^\star)$. Then, we get $\sum c_i r(\phi_i(x))s_i(y^\star)$. Suppose $r(t)=\exp(t)$. Then, we get $\sum \exp(d_i^\star + \phi_i(x))$ for $d_i^\star$ that is constant in $y^\star$ and the coefficients. Then, we get that $\log \partial_y T(x,y^\star)$ is really a log-sum-exp of $\phi_i + d_i^\star$, which people seem to like :grin:

mparno: The $r(\prod\phi)$ vs $\prod r(\phi)$ was where my concern was coming from and I just misread the code.  Sorry Danny, but thanks for clearing that up!

mparno (approved):
Thanks for clearing things up for me Danny.  This looks good to me now.

MParT/RectifiedMultivariateExpansion.h
dannys4: resolved?

PR #421, opened 2024-05-14
Coefficient Bounds

### LINKED ISSUE #419: Add interface for constraints on parameters
Unlike the `MonotoneComponent` class, the `RectifiedMultivariateExpansion` class requires some of the expansion parameters to be positive in order for the resulting map to be monotone.  We currently don't have a good way of defining or imposing these bounds during map training.    I propose we add a `ParameterBounds()` function to the `ConditionalMapBase` class that returns any lower or upper bounds on the map parameters that are needed to ensure monotonicity.  The default behavior could be [-inf,inf] indicating that no bounds are necessary.

mparno:
Closes #419 .   

This PR introduces the `CoeffBounds` function to the `ConditionalMapBase` interface that returns the lower and upper bounds on the map coefficients that guaranteee the map is monotone.


dannys4 (commented):
This doesn't seem to actually work on Julia. I'll put up a fix tomorrow on this and the MParT.jl repo, then I can approve.

dannys4 (approved):
Looks great!

PR #427, opened 2025-11-04
Fix PyPI and kokkos on m1

dannys4:
- We didn't constrain versions of ninja and cmake which has become a problem if installing via pip.
- some issue happens now for kokkos <4.5 so I just changed the default version in cmake.


dannys4:
sorry for auto linting

mparno (approved):
Thanks Danny!  Sorry for the slow response on my end.

